PR #48, opened 2019-02-11
Reorganize text steps

eriksiegel:
A relatively larger number of changes around text steps

- Added the text related steps `text-count, -head, -join, -replace, -sort, -tail` to the *standard* step library. We talked about this and weren't totally clear on it in Prague but IMHO these steps are pretty basic and should be in the standard library. Feel free to disagree. Maybe text-sort should go to optional?
- The optional text steps now contain only one: `markdown-to-html`. Its at this moment just a skeleton description, Norm volunteered to write its description since he already created such a step.
- `text-replace` is just a wrapper around `fn:replace()`. That might seem unnecessary, but I think its worthwhile to add such a step since that's what you're probably going to do a lot when working with text. Having it in a standard XProc step makes the pipelines better readable.
- `text-sort` is modeled on XSLT's `<xsl:sort>`. I took all its attributes and turned them in to step options. Too much?


ndw:
I think the markdown step should be called `p:markdown`. The -to-html part is redundant, I think. Otherwise, I'm happy to put these in and review them in detail later.

ndw:
You didn't ask for reviewers, so I can't officially approve this PR, but consider it approved anyway 😄

PR #55, opened 2019-03-21
Skeleton archive unarchive steps

eriksiegel:
I created skeleton entries for the p:archive and the p:unarchive steps. I suggest we use these as a starting point for further refinements (even if they're not complete yet), unless I of course totally interpreted things wrong. And even then, something to start with and bicker about is better than nothing.

This is for #3 en #4, based on my interpretation of the minutes of Prague, https://github.com/xproc/Workshop-2019-02/wiki/Meeting-minutes-Wednesday

The specifications need a schema extension for their manifest file(s) but that is currently beyond my abilities.

### DIFF
--- a/src/main/xml/bibliography.xml
+++ b/src/main/xml/bibliography.xml
@@ -329,5 +329,7 @@ Internet Engineering Task Force. July, 2005.</bibliomixed>
 459</citetitle>. <biblioid class="doi">10.1109/DSN.2002.1028931</biblioid>.
 P. Koopman. June 2002.
 </bibliomixed>
-
+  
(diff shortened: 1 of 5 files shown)


xml-project:
Thanks Erik. I haven't had the time to read it all, but I will do tomorrow morning. On the first reading of p:unarchive two question came up:

1. Do we really need option `format`? Wouldn't it be enough to say, the implementation has to choose the method based on the source document's content type? May be I overlooked something obvious.

2. I would like to have the manifest either as XML or as JSON. If I remember right, we talked about this in Prague, so the mechanism to select this would, given the current signature, one special key in `parameters`? Right?

PR #56, opened 2019-03-25
First shot at json-join

xml-project:


xml-project:
I renamed the step from join-json  to json-join, so we are inline with the steps created for text, e.g. text-join.

xml-project:
> Otherwise Ok, unless one also wants to cram map:merge() functionality into this function. Or will there be another p:merge-json step for merging by common key?

Yes, you are right. We should first decide, which XPath function for JSON we wrap into XProc steps and then I will come back to this. Topic for next call?

Issue #57, opened 2019-04-08
Outdated note in p:directory-list?


xml-project:
In [p:directory-list](http://spec.xproc.org/master/head/steps/#c.directory-list) there is a note saying:

> There is no way to specify a list of values using attribute value templates. If the option shortcut syntax is used to provide the include-filter or exclude-filter option, it will consist of a single regular expression. To specify a list of regular expressions, you must use the p:with-option syntax.

Since we now `include-filter` and `exclude-filter` are defined as `xs:string` (whitespace separated lists of regular expressions) I fail to understand what "list" is supposed to mean in this context. Is this box a remnant from the time, we played with `xs:string*`?

eriksiegel:
Ah yes. Data type of both options should be `xs:string*`

xml-project:
@eriksiegel Sure? I thought we opted against it and use a whitespace separated string. (The text says so)

gimsieke:
At least the [change](https://github.com/xproc/3.0-specification/blame/f04a90106807441ddabbf9b6ccfa819905b919bd/steps/src/main/xml/steps/directory-list.xml#L39) where we went from a list of regexes to whitespace-separated regexes is more recent (Jan 29, 2019) than the note that you cited (it’s from Aug 19, 2018).

When looking at the [PR that introduced the change](https://github.com/xproc/3.0-specification/pull/732), I don’t see an approval or a reference to an issue. 

I don’t remember what we decided. The London 2018-06 notes seem to suggest that we went for lists of regexes, and that’s what the Aug 19 notes refers to.

I’m not against space-separated regexes since URIs may not contain whitespace anyway. 

But I’m wondering whether the circumstances that made Norm change this option from “sequence” to “space-separated” also affected other options. The options `include-filter` and `exclude-filter` of `p:load-directory-list` and the attribute `namespace` on `p:import-functions` seem to be affected by the change.

xml-project:
> But I’m wondering whether the circumstances that made Norm change this option from “sequence” to “space-separated” also affected other options. The options include-filter and exclude-filter of p:load-directory-list and the attribute namespace on p:import-functions seem to be affected by the change.

At least `p:unarchived` seems to be affected because there we introduced space separated lists too. May be @ndw can help.

In general I would have a slight tendency for the "list of strings" (xs:string*), but obviously it was decided otherwise.

eriksiegel:
[...]

PR #62, opened 2019-04-22
Attempt to fix 58

### LINKED ISSUE #58: Description of p:directory-list needs improvement
In p:directory-list there is no hint, that the value of option `path` has to be resolved against the relevant baseURI if it is relative. In XProc 1.0 this hint was missing too, but the test in the test suite required this behavior.
I think we should make this explicit.

xml-project:
Attempt to fix #58


xml-project:
Also related #57

PR #65, opened 2019-04-23
attempt to improve p:archive (archive input port, manifestless creation, …)

gimsieke:
Also provide an RNC schema for the manifest.

I renamed Erik’s `c:file` back to to `pxp:zip`’s `c:entry` so that we don’t have to differentiate between `c:file` in directory listings and `c:file` in archive entries.

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -774,6 +774,14 @@ AnyElement =
    }
 
 [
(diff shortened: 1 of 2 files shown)


xml-project:
@gimsieke Close this?

gimsieke:
Sure, but how?

Issue #54, opened 2019-03-21
Move p:directory list?


eriksiegel:
Shouldn't p:directory-list be moved to the file steps?
The reason the file steps are optional (not everything has a file system) also applies to p:directory-list.

xml-project:
+1

ndw:
Achim, you've got another issue to edit one of the directory-list steps. If I move them while you're editing them, we'll just get a nasty merge conflict. Can you move them when you're making your other changes?

xml-project:
OK, I see the problem. 

Have we finally decided to move p:directory-list (and p:load-directory-list)? Missed that.

> Can you move them when you're making your other changes? 

What changes do I have to make on p:directory-list? Though I am done.

ndw:
I thought we had consensus on that, but if you disagree...

Anyway #59 is still open and assigned to you, so I assumed you were working on something.

xml-project:
Nope, I agree - just lost the consensus.

> Anyway #59 is still open and assigned to you, so I assumed you were working on something.

OK, I thought you referred to required changes on p:directory-list which I left out.

Will move both steps to the file-steps.

eriksiegel:
Done with #212

PR #66, opened 2019-04-24
Fix 59

### LINKED ISSUE #59: p:load-directory-list's restriction to application/xml
Currently p:load-directory-list requires "application/xml" to appear on result port. I think we do not need this restriction and load other types of documents too.

Where is my mistake?

xml-project:
Attempt to fix #59  and #54


xml-project:
@ndw @gimsieke 
Did you miss this?

step-file/src/main/xml/specification.xml
```diff
@@ -62,6 +62,169 @@ steps is assumed; for background details, see
...
+          <tag>c:other</tag> element but the criteria for marking a file as
+          special are <glossterm>implementation-defined</glossterm>.</impl>
+        </para>
+        <e:rng-pattern name="VocabOther"/>
+      </listitem>
+    </itemizedlist>
+    
+    <para>When a directory entry is a subdirectory, that directory's entries are not
+      output as part of that entry's <tag>c:directory</tag>.  A user must apply this step
+      again to the subdirectory to list subdirectory contents.</para>
```
gimsieke: Didn’t we want to have a `recursive` option?

xml-project:
Hi @gimsieke ,
thanks for taking the time. I just tried to solve the two issues, but you are right. We should have a consensus on the mentioned issues first. Would you mind raising the issues, so we have the points in mind.

PR #68, opened 2019-04-26
Changed content-type of ports to short cuts

xml-project:


xml-project:
@ndw @gimsieke @eriksiegel 
I changed the signature of 37 steps in the standard lib to content-type shortcuts.
Could somebody please have to look at this **asap** to avoid merging problems with other changes.

PR #81, opened 2019-05-02
Attempt to fix 78

### LINKED ISSUE #78: Content-type of source in p:hash is wrong
The source port is flagged as "any" but it has to be "XML HTML" given that the NODE where the hash is inserted is selected by a MATCH-PATTERN.

xml-project:
Trying to fix #78 
I changed to signature (source = xml html / result = xml html) and added an error for a replacement like "/" resulting in a single text node. 
We might allow this but then it is a new (text) document and we can no longer say, that all document properties are preserved (for the content-type changed!)

### DIFF
--- a/steps/src/main/xml/steps/hash.xml
+++ b/steps/src/main/xml/steps/hash.xml
@@ -9,8 +9,8 @@
 for some value and injects it into the <port>source</port> document.</para>
 
 <p:declare-step type="p:hash">
-  <p:input port="source" primary="true" content-types="any"/>
-  <p:output port="result" content-types="application/xml"/>
+  <p:input port="source" primary="true" content-types="xml html"/>
+  <p:output port="result" content-types="xml html"/>
   <p:option name="parameters" as="map(xs:QName,item()*)?"/>
   <p:option name="value" required="true" as="xs:string"/>
   <p:option name="algorithm" required="true" as="xs:QName"/>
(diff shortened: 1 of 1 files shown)


xml-project:
1. Yes, you are right. We have to change the default value for match. Your proposal seems reasonable. Will change that.
2. I think the problem is, that with @select on p:input etc. we say, that multiple text node are combined to one and that a document node is wrap around it and that it is a text document. If we say so, p:hash may change the content-type and therefore no document-properties are preserved. In most cases I think someone will add a hash to an element or an attribute and expected the document-properties to be preserved. If we guess so, we need to say, that a text only result of p:hash is an error.

gimsieke:
I saw this: http://spec.xproc.org/master/head/xproc/#p.input.6.4.2.1.6
> If such a document consists exclusively of text nodes, then it has a content type of text/plain, …

and this immediately before http://spec.xproc.org/master/head/xproc/#err.inline.D0016.1
> If a document constructed by selection consists exclusively of text nodes, then it has a content type of text/plain, …

But here (http://spec.xproc.org/master/head/xproc/#documents.9) we say:
> If the result is a single text node, it is wrapped in a document node and the document's content-type is text/plain.

We don’t say anything about a result that consists of multiple text nodes. Maybe we tacitly assumed that [sequence normalization](https://www.w3.org/XML/Group/qtspecs/specifications/xslt-xquery-serialization-31/html/Overview.html#sequence-normalization) is applied to step results and that multiple text nodes therefore were combined to a single text node. We should state this explicitly in the spec.

xml-project:
@gimsieke Yes, we should fix this. Would you mind to raise an issue for the main specs.
Do we have to fix this before we say that p:hash should raise an error if it produces a text node only document?

xml-project:
I changed the default value for match as proposed and the error text, so that joining text nodes is no longer implied.

xml-project:
@gimsieke @ndw We could also remove the default value for `match` on `p:hash` and make the option required? Would make sense to me!

xml-project:
@ndw 
@gimsieke approved this pull request but asks for your approval. If you have time, would you mind to have a look at it?

PR #83, opened 2019-05-03
Attempt to fix 75

### LINKED ISSUE #75: Preserved document properties of p:escape-markup
Currently for [p:escape-markup](http://spec.xproc.org/master/head/steps/#c.escape-markup) it is said, that no document properties are preserved.

One could understand this step as a special kind of p:insert (which preserves all properties), so I would expected p:escape-markup to do this too. 

Did I miss something?

xml-project:
Trying to fix #75

### DIFF
--- a/steps/src/main/xml/steps/escape-markup.xml
+++ b/steps/src/main/xml/steps/escape-markup.xml
@@ -38,6 +38,6 @@ XML declaration in the escaped result.</para>
 
 <simplesect>
 <title>Document properties</title>
-<para feature="escape-markup-preserves-none">No document properties are preserved.</para>
+<para feature="hash-preserves-all">All document properties are preserved.</para>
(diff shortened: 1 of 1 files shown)


steps/src/main/xml/steps/escape-markup.xml
```diff
@@ -38,6 +38,6 @@ XML declaration in the escaped result.</para>
 
 <simplesect>
 <title>Document properties</title>
-<para feature="escape-markup-preserves-none">No document properties are preserved.</para>
+<para feature="hash-preserves-all">All document properties are preserved.</para>
```
gimsieke: I don’t know much about what goes into the feature attribute, but `hash-preserves-all` on `p:escape-markup` looks dubious. I would have expected `escape-markup-preserves-all`.

PR #98, opened 2019-05-25
Added @uri to c:file etc.

xml-project:
Added attribute `uri`, so the complete uri (and not just the name) is included in c:file etc. returned by p:directory-list.
Could somebody please confirm this PR soon, because we need this for our London tutorial.

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -818,6 +818,7 @@ VocabParamSet =
 VocabDirectory =
    element c:directory {
       attribute name { text },
+      attribute uri { text },
       ((VocabFile|VocabDirectory|VocabOther)*)
    }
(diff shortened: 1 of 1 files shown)


xml-project:
As @gimsieke point out, he wants another approach, which he has to specify first. Therefore I have to find another solution for my problem.

gimsieke:
I finally agree with `@uri`. 
My concerns wrt redundancy can be addressed: Implementations may (but are not required to) use relative URIs. If `@uri` may be relative, the pipeline author needs to make sure to resolve it against the base URI though (which is always given on the top-level `c:directory` element). Or do you want to require an implementation to attach absolute URIs?

Please don’t close pull requests prematurely. Discussions are a part of the process.

Issue #116, opened 2019-06-11
Clarify p:archive and p:unarchive


ndw:
1. It's an error if two documents with the same base URI appear on the source port
2. It is an error if a document without a base URI appears on the source port
3. It is an error if more than one document appears on the archive port
4. ...

eriksiegel:
* unarchive should no longer do any manifest generation
* filters must be string*

eriksiegel:
As of 20190620:

* No content directly in the manifest
* Serialization document property is used

eriksiegel:
Closed with one thing to discuss left: #155

eriksiegel:
Looking at what is already done, I think this one can be closed.

Issue #118, opened 2019-06-12
Incorrect definition for archive manifest c:entry


eriksiegel:
The definition for a `c:entry` element in an archive's manifest now reads `anyElement*`. This should be `anyNode*` (does something like that exists? Maybe just mixed contents with any elements?).

If we keep it `anyElement*` you will never be able to specify text contents in a manifest, which is rather handy!

gimsieke:
You can, if you wrap it in an element (that in turn may have any child nodes I guess, at least this was the premise on which I allowed almost any child element)

eriksiegel:
How do we then make the distinction between an entry with just text and an entry with a single root element? There is no content-type attribute...

gimsieke:
I think I see what you mean here. Your assumption probably is that `c:entry` can hold the actual contents of an archive member. This is not the case. 
Please correct me if my interpretation of your question is wrong.

When you create a manifest, any `c:entry` child nodes will be ignored (as will be specified once we fix “When the c:entry elements has any child nodes, it is ignored.” → “When the c:entry elements has any child nodes, they are ignored.”). The archive contents will be pulled from the `source` port or, if there’s no document with a matching base URI, from the location given by `c:entry/@href`.

In the manifest that comes out on the `report` port of `p:archive`, `c:entry` wasn’t meant to hold the actual content of an archive entry, either. We allowed children in `c:entry` in order to allow archiving tools to enrich entries with additional metadata or error reporting that isn’t covered by the existing attributes such as `@level`, `@comment`, `@method`, and implementation-defined attributes.

eriksiegel:
There seems to be a difference of opinion here. What I meant is (and probably said wrong in London):
When the c:entry elements has any child nodes, it is ignored. ==> When the c:entry elements has any child nodes, the **href attribute** is ignored

So what I would like is that the `p:archive` steps looks for the contents of the archive entries in three ways (in this order!):

1. If `c:entry` has any contents, this will become the contents of the entry. `c:entry/@href` is ignored.
1. If the `href` attribute has the same value as the base URI of one of the document appearing on the `source` port, this is used
1. The `href` attribute is interpreted as a URI and the document is loaded from disk or whatever.

We can change the mechanism, but for some of my use-cases it is rather important that you can specify archive entry contents in the manifest. I do would like to have something that enables that.

eriksiegel:
Closed because of decision on #116

PR #119, opened 2019-06-12
116 archive steps

eriksiegel:
Changes related to #116.

I almost completely rewrote the `p:archive` section (without hopefully breaking any intended meaning) because in my opinion it was not clear and not strict enough. So @gimsieke, please read it carefully.

I think it will be rather problematic to verify all changes without seeing the resulting HTML (and I didn't yet master the art of creating GitHub pages for my forked repos). We could merge the PR and take it from there. 

There is still one editorial note that I have no answer for. To be discussed in a call I think.

### DIFF
--- a/steps/src/main/xml/steps/archive.xml
+++ b/steps/src/main/xml/steps/archive.xml
@@ -4,12 +4,13 @@
 
   <title>p:archive</title>
 
-  <para>The <code>p:archive</code> step outputs on its <port>result</port> port an archive (usually binary) document,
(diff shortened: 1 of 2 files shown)


eriksiegel:
There is still discussion #118 going on about this...

eriksiegel:
@ndw @xml-project @gimsieke May I suggest we merge this one? Its a very hard change to approve without seeing the HTML. I think there are still a couple of loose ends and maybe disagreements, but I'm perfectly happy to edit it again, after discussions. 

Maybe put it on the agenda for the 20th?

PR #120, opened 2019-06-13
Attempt to fix 102

### LINKED ISSUE #102: Output port of p:text-join needs a fix
Currently port "result" of [p:text-join](http://spec.xproc.org/master/head/steps/#c.text-join) has `content-type`= "text". Since a new document is created here, I think we need to define the media type correctly.
Proposal: text/plain.

xml-project:
Attempt to fix #102


xml-project:
@gimsieke 
Added the requested change. As the modified error is also raised by p:cast-content-type, I changed this too.
Please approve my changes.

PR #125, opened 2019-06-14
Attempt to fix 122

### LINKED ISSUE #122: p:delete deleting the document element
In the light of #100 and #93 I propose to adapt this change to p:delete too: If the resulting document is a document node containing just one text node, the content-type should switch to "text/plain" and the serialization entry in the document properties should be deleted.

xml-project:
Attempt to fix #122


xml-project:
Come back to this later,

PR #127, opened 2019-06-15
remove uri attribute on c:directory; correct xml:base in example

gimsieke:
Not sure whether the change in src/main/schema/core30.rnc will immediately show in the generated syntax boxes. We might need to update the same file in the 3.0-specification repo.

### DIFF
--- a/src/main/schema/core30.rnc
+++ b/src/main/schema/core30.rnc
@@ -818,7 +818,6 @@ VocabParamSet =
 VocabDirectory =
    element c:directory {
       attribute name { text },
-      attribute uri { xsd:anyURI },
       xmlbase.attr,
       file-info.attributes*,
       ((VocabFile|VocabDirectory|VocabOther)*)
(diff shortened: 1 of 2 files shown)


gimsieke:
(only asking @xml-project for approval because he raised both issues in an email message earlier today)

gimsieke:
That I didn’t remove `@uri` on the other two elements was just an oversight.

Hm, `c:file/@content-types` was introduced by @ndw in 305e72ec (which was just the commit in which he moved the steps in a separate repo). It originates from https://github.com/xproc/3.0-specification/commit/831031cf8302b97b5d99a127b07c1e69847310d4 where he said that he wanted to add a `@content-type` (no plural) attribute to `c:file`. 

I suggest that if we want to allow `@content-type` on `c:file` we should allow it as part of the [detailed result attributes](https://github.com/xproc/3.0-steps/commit/3618dc1c3c6746a392bdfa0b084d2acbff48fac4#diff-240748684ef21d0af7e83fd762195207R849). These should be the same attributes that are available on the result of [`p:file-info`](http://spec.xproc.org/master/head/file/#c.file-info). But I’m not sure whether determining the content type is as easy and reliable as finding out whether a file is writable, or finding out its size. This needs to be discussed in a separate issue I guess.

xml-project:
👍

PR #131, opened 2019-06-16
Fix unescape markup

xml-project:
Removed the notion of "document element", so all element childs of the document node are now unescaped.

### DIFF
--- a/steps/src/main/xml/steps/unescape-markup.xml
+++ b/steps/src/main/xml/steps/unescape-markup.xml
@@ -6,9 +6,9 @@
 <title>p:unescape-markup</title>
 
 <para>The <tag>p:unescape-markup</tag> step takes the string value of
-the document element and parses the content as if it was a Unicode
+the document's child elements and parses the content as if it was a Unicode
 character stream containing serialized XML. The output consists of the
-same document element with children that result from the parse. This
+same elements with children that result from the parse. This
 is the reverse of the <tag>p:escape-markup</tag> step.</para>
 
 <p:declare-step type="p:unescape-markup">


xml-project:
I agree with (2).

> Can we keep the sentence “When the string value is parsed, the original document element is preserved so that the result will be well-formed XML even if the content consists of multiple, sibling elements.” if there can be multiple top-level elements?

What do you mean by "keep the sentence"? I could not find this sentence in the current specs, so I did not remove it. I also fail to see the rational behind it. Consider the following document on source
````
<element>escaped text</element>
<element>more escaped text</element>
````
If there is no well-formed XML on source, how to produce it on result by unescaping its children? 
What did I miss?

My central reason for proposing this change is, that p:unescape-markup is the reserve for p:escape-markup.

gimsieke:
> 
> 
> I agree with (2).
> 
> > Can we keep the sentence “When the string value is parsed, the original document element is preserved so that the result will be well-formed XML even if the content consists of multiple, sibling elements.” if there can be multiple top-level elements?
> 
> What do you mean by "keep the sentence"? I could not find this sentence in the current specs, so I did not remove it. I also fail to see the rational behind it. Consider the following document on source

https://github.com/xproc/3.0-steps/blob/a33717a2f5b7724fedaa52fb7bd60b7b984d4639/steps/src/main/xml/steps/unescape-markup.xml#L28

> 
> ```
> <element>escaped text</element>
> <element>more escaped text</element>
> ```
> 
> If there is no well-formed XML on source, how to produce it on result by unescaping its children?
> What did I miss?
> 
> My central reason for proposing this change is, that p:unescape-markup is the reserve for p:escape-markup.

I wonder whether there’s value in having `p:escape-markup` optionally producing a text document. Optionally because of backward compatibility. If we’d devise this as a new step in XProc 3.0, we’d probably go with a text document for the escaped markup, with no elements around it.

xml-project:
So the requested sentence is still there. What do I have to change so you will approve this RP?

xml-project:
Ok, now I got it. Sorry for being confused. I understand that you wanted me to keep (=not remove) the sentence.

Yes, you are right. This sentence is a problem for XDM documents with multiple top-level elements.
Actually: I do not really understand, what this sentence means for well-formed xml.

A topic for Thursday's call, I guess.

gimsieke:
I propose that we delete the paragraph “When the string value is parsed […] multiple, sibling elements.” since we abandoned the concept of a single “document element” and because the result, when serialized, will not be well-formed if the input already contained multiple top-level elements.

xml-project:
Copied your proposal to issue #137

Issue #115, opened 2019-06-11
Add a p:archive-manifest step


ndw:
Given an archive, it returns an XML document that is the manifest for the current contents of that archive.

eriksiegel:
The unarchive step also mentions returning a manifest. That's no longer true.

PR #140, opened 2019-06-19
Added p:archive-manifest

eriksiegel:
See #115

### DIFF
--- a/steps/src/main/xml/steps/archive-manifest.xml
+++ b/steps/src/main/xml/steps/archive-manifest.xml
@@ -1,0 +1,45 @@
+<section xmlns="http://docbook.org/ns/docbook" xmlns:p="http://www.w3.org/ns/xproc"
+  xmlns:e="http://www.w3.org/1999/XSL/Spec/ElementSyntax" xmlns:xi="http://www.w3.org/2001/XInclude"
+  xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="c.archive-manifest">
+
+  <title>p:archive-manifest</title>
+
+  <para>The <code>p:archive-manifest</code> creates an XML manifest file describing the contents of
+    the archive appearing on its <port>source</port> port.</para>
+
+  <p:declare-step type="p:archive-manifest">
+    <p:input port="source" primary="true" content-types="*/*" sequence="false"/>
+    <p:output port="result" primary="true" content-types="application/xml" sequence="false"/>
+    <p:option name="format" as="xs:QName" required="false" select="'zip'"/>
+    <p:option name="parameters" as="map(xs:Qname, item()*)" required="false"/>
+  </p:declare-step>
+
+  <para>The <code>p:archive-manifest</code> step inspects the archive appearing on its
+      <port>source</port> port and outputs a manifest describing the contents of the archive on its
+      <port>result</port> port. </para>
+
+  <para>The format of the archive can be specified using the <option>format</option> option.
+    Implementations <rfc2119>must</rfc2119> support the <biblioref linkend="zip"/> format, specified
+    with the value <code>zip</code>. <impl>It is <glossterm>implementation-defined</glossterm> what
+      other formats are supported.</impl></para>
+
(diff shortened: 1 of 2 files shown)


eriksiegel:
If we do this, the same mechanism should apply to p:archive and p:unarchive...???

Honestly, I don't think its worth the burden on the implementers. Sounds like a very fringe case to me. Any opinions?

xml-project:
The phrase "implementations should try to infer the archive format from the source document’s content-type property" is vague enough in my opinion to hint on a quality of implementation. I would prefer to have this phrase, because if it is missing, the implementation might be *required* to use the default value, even if it knows what to do with "application/zip".

eriksiegel:
> the implementation might be required to use the default value, even if it knows what to do with "application/zip"

But the default *is* `application/zip` (zip is the default value for the format option). I'm not sure what you mean

eriksiegel:
I tried to process the comments made at our last editorial meeting in the p:archive-manifest description. Please have a look.

If this is ok, I'll revisit and rework the p:archive/unarchive likewise.

steps/src/main/xml/steps/archive-manifest.xml
```diff
@@ -0,0 +1,64 @@
...
+        supported.</impl> <error code="C0081">It is a <glossterm>dynamic error</glossterm> if the format of the
+          archive does not match the format as specified in the <option>format</option>
+          option.</error></para>
+    </listitem>
+    <listitem>
+      <para>If no <option>format</option> option is specified or its value is the empty sequence,
+        the archive's format will be determined by the step, using the <code>content-type</code>
+        document-property of the document on the <port>source</port> port and/or by inspecting its
+        contents. <impl>It is <glossterm>implementation-defined</glossterm> how the step determines
+          the archive's format.</impl> Implementations <rfc2119>should</rfc2119> recognize archive's
```
gimsieke: recognize archive's → recognize archives

steps/src/main/xml/steps/archive-manifest.xml
```diff
@@ -0,0 +1,64 @@
...
+      <para>If the <option>format</option> option is specified, this determines the format of the
+        archive. Implementations <rfc2119>must</rfc2119> support the <biblioref linkend="zip"/>
+        format, specified with the value <code>zip</code>. <impl>It is
+            <glossterm>implementation-defined</glossterm> what other formats are
+        supported.</impl> <error code="C0081">It is a <glossterm>dynamic error</glossterm> if the format of the
+          archive does not match the format as specified in the <option>format</option>
+          option.</error></para>
+    </listitem>
+    <listitem>
+      <para>If no <option>format</option> option is specified or its value is the empty sequence,
```
gimsieke: or its value → or if its value
(I’m not too confident about this change, but it sounds better to me.)

steps/src/main/xml/steps/archive-manifest.xml
```diff
@@ -0,0 +1,64 @@
...
+    archive cannot be understood, determined and/or processed.</error></para>
+
+  <para>The <option>parameters</option> option can be used to supply parameters to control the
+    archive manifest generation. <impl>The semantics of the keys and the allowed values for these
+      keys are <glossterm>implementation-defined</glossterm>.</impl>
+    <error code="C0079">It is a <glossterm>dynamic error</glossterm> if the map
+        <option>parameters</option> contains an entry whose key is defined by the implementation and
+      whose value is not valid for that key.</error></para>
+
+  <para>The generated archive has the format as described in <xref linkend="cv.archive-manifest"/>.
```
gimsieke: generated archive → result document

eriksiegel: ok, done

gimsieke: The sentence still seems to read “The generated archive has the format as described in …”. But there is no generated archive. You can call it “generated archive manifest” or, as I proposed, “result document”.

eriksiegel:
Sorry, forgot this, will fix.

PR #146, opened 2019-06-22
Attempt to fix 94

### LINKED ISSUE #94: p:set-attributes need an update due to change in document model
Currently the [specs](http://spec.xproc.org/master/head/steps/#c.set-attributes) use the concept of a document element. Since we use XDM now, we have three options (I think) to update this step:

1. We take the attributes from the _first_ element child of the document-node.
2. We take the attributes from _all_ element children of the document-node.
3. It is an error if the document on port `attributes` has more than one child element.

Do you see other options? Which one would you prefer?

xml-project:
Attempt to fix #94 
Port "attributes" removed and using map as option instead.

### DIFF
--- a/steps/src/main/xml/steps/set-attributes.xml
+++ b/steps/src/main/xml/steps/set-attributes.xml
@@ -12,7 +12,7 @@ matching elements.</para>
    <p:input port="source" primary="true" content-types="xml html"/>
    <p:output port="result" content-types="xml html"/>
    <p:option name="match" as="xs:string" select="'/*'" e:type="XSLTSelectionPattern"/>
-   <p:option name="attributes" required="true" as="map(xs:QName, xs:string)" /> 
+   <p:option name="attributes" required="true" as="map(xs:QName, xs:untypedAtomic)" /> 
 </p:declare-step>
 
  <para>The value of the <option>match</option> option <rfc2119>must</rfc2119> be an
@@ -22,8 +22,8 @@ matching elements.</para>
 
 <para>A new attribute is created for each entry in the map appearing
 on the <option>attributes</option> option. The attribute name is taken
-from the entry's key while the attribute value is taken from the entry's
-value.</para>
+from the entry's key while the attribute value is taken from the string value of
+the entry's value.</para>
 
 <para>If an attribute with the same name as one of the attributes to
 be created already exists, the value specified on the
--- a/steps/src/main/xml/steps/set-attributes.xml
+++ b/steps/src/main/xml/steps/set-attributes.xml
@@ -10,9 +10,9 @@ matching elements.</para>
 
 <p:declare-step type="p:set-attributes">
    <p:input port="source" primary="true" content-types="xml html"/>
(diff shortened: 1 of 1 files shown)


steps/src/main/xml/steps/set-attributes.xml
```diff
@@ -10,33 +10,33 @@ matching elements.</para>
 
 <p:declare-step type="p:set-attributes">
    <p:input port="source" primary="true" content-types="xml html"/>
-   <p:input port="attributes" content-types="xml html"/>
    <p:output port="result" content-types="xml html"/>
    <p:option name="match" as="xs:string" select="'/*'" e:type="XSLTSelectionPattern"/>
+   <p:option name="attributes" required="true" as="map(xs:QName, xs:string)" />
```
gimsieke: Shouldn’t we allow, in principle, typed attributes, and therefore accept `map(xs:QName, xs:anyAtomicType)` here?

xml-project: Fine with me. We then have to say that the value of the created attribute is the string value of the map entry's value, right?

gimsieke: I’m not sure whether typed attributes can only be created when PSVI support is enabled and when the type annotations are conveyed in the PSVI. But then a step is probably not allowed to create such an annotation, because of 
> In order to avoid these inconsistencies, most steps must not produce PSVI annotated results even when PSVI passing is supported.

(see http://spec.xproc.org/master/head/xproc/#psvi-support). Maybe @ndw can chime in here and enlighten us whether steps may create typed attributes at all. 
If not, we might as well only accept string values. I suppose that if people supply, for ex., integers as map values, they’d be typecast to string then without raising an error?

xml-project: 👍

ndw: Constructed attributes in XSLT and XQuery are always of type `xs:untypedAtomic`. I think that's what we should say. The string value as an untyped atomic. (That's good enough for a lot of cases because, for example, if you attempt to add 3 to an untyped atomic, the processor will attempt to make the untyped atomic into a number and will succeed if it's numeric. If we make it of type xs:string, then that doesn't work.

steps/src/main/xml/steps/set-attributes.xml
```diff
@@ -10,33 +10,33 @@ matching elements.</para>
...
-appears on the <port>attributes</port> port is copied to each element
-that matches the <option>match</option> expression.</para>
+<para>A new attribute is created for each entry in the map appearing
+on the <option>attributes</option> option. The attribute name is taken
+from the entry's key while the attribute value is taken from the entry's
+value.</para>
 
 <para>If an attribute with the same name as one of the attributes to
-be copied already exists, the value specified on the
-<port>attribute</port> port's document is used. The result port of
```
ndw: Aren't you missing the words "parameters map" or something like that here?

PR #149, opened 2019-06-27
Updated p:unarchive to decisions made in last call

eriksiegel:
This implements #116 for the decisions we made in our last call about `p:unarchive` 

I made a decision about the base URI of unarchived documents which we did not discuss, might be controversial, please comment: 

>The base URI of an unarchived document appearing on the result port is the relative path of this document as it was in the archive. For instance, the base URI of an unarchived file called xyz.xml that resided in the specs subdirectory in the archive will become specs/xyz.xml.

### DIFF
--- a/steps/src/main/xml/steps/unarchive.xml
+++ b/steps/src/main/xml/steps/unarchive.xml
@@ -4,34 +4,48 @@
 
   <title>p:unarchive</title>
 
-  <para>The <code>p:unarchive</code> step outputs on its <port>result</port> port specific
-    entries in an archive (for instance from a zip file).</para>
+  <para>The <code>p:unarchive</code> step outputs on its <port>result</port> port specific entries
+    in an archive (for instance from a zip file).</para>
 
   <p:declare-step type="p:unarchive">
     <p:input port="source" primary="true" content-types="*/*" sequence="false"/>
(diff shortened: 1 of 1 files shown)


steps/src/main/xml/steps/unarchive.xml
```diff
@@ -64,10 +78,11 @@
...
     values.</para>
-  
-  <note role="editorial">
-    <para>What about the base URIs of these documents?</para>
-  </note>
+
+  <para>The base URI of an unarchived document appearing on the <port>result</port> port is the
+    relative path of this document as it was in the archive. For instance, the base URI of an
+    unarchived file called <code>xyz.xml</code> that resided in the <code>specs</code> subdirectory
+    in the archive will become <code>specs/xyz.xml</code>.</para>
```
gimsieke: Should/may the top-level element have an absolute base URI? Is it the URI of the archive? If it is the URI of a file, this will be problematic for the relative base URIs of the entries (they will resolve to the directory that the archive resides in, if it resides in a directory). 
In order to make the manifest that is returned by this step immediately useful as an input manifest for `p:archive`, the extract directory’s URI should be the top-level element’s base URI. But we left out the possibility of letting `p:unarchive` immediately store the files on disk. Should we revise this decision by allowing an optional `href` or `output-directory` option?

PR #154, opened 2019-06-27
104 pxquery update

eriksiegel:
See #104. Tried to make the description of `p:xquery` analogues to how we did `p:xslt`


xml-project:
@eriksiegel Would you mind to check this PR. For me it contains the changes to p:os-info and just deletions for p:query (deletion of lib-ref and a paragraph).

eriksiegel:
Sorry, as @xml-project noticed, made a mistake and did too much here by accident.

Issue #165, opened 2019-07-16
Add default values for options that are maps


ndw:
We have several steps with map-type options, e.g. serialization.

Currently, they don't specify a default value so they get the default value `()`. That doesn't work. I think the easiest solution is to give them all default values of `map { }`.

xml-project:
👍 
But isn't that a step issue as only option default to `()`? Did I miss the passage where we say, that attributes like `p:output/@p:serialization` also default to `()`?

ndw:
It is a step issue, the particular step I was implementing was `p:escape-markup` :-)

I don't think we have to do or say anything about maps in places like `p:output` because they're part of the processor.

xml-project:
Sorry, I though you raised that error in the core specs repo.
Also: I was surprised to learn that `() instance of map(*)` is false. But it is!

gimsieke:
Maybe we can specify map options as `"map(xs:Qname, item()*)?` instead of `"map(xs:Qname, item()*)`, then `()` is a valid default

ndw:
That would make empty sequence valid, but it would also mean that every step implementation has to check if the value it received is an empty sequence. I think it's simpler if the default is an empty map.

xml-project:
It seems to me, that the solution proposed by @gimsieke is widely used in the step library currently. But you really think we need to change that?

ndw:
🤷‍♂ Fine by me, I suppose

xml-project:
@ndw Did I miss something?

ndw:
It's fine. It's just slightly more work in the implementation of each step because you have to check for empty sequence instead of having an empty map provided automatically. But it's probably more consistent and simpler for users, so I'm fine with it. I wrote a convenience method 😄

ndw:
I think this is closed by #171 as well.

PR #171, opened 2019-07-24
Cleanup optional option types

### LINKED ISSUE #170: The 'override-content-type' option on p:text-join should be optional
The type should be `xs:string?` otherwise an empty sequence is either an error or indistinguishable from an empty string, I'm not sure which.

ndw:
Fix #170


ndw:
Also fix #165

PR #175, opened 2019-07-25
Fix several simple, editorial issues

### LINKED ISSUE #110: Move p:json-join to the standard step library.

### LINKED ISSUE #166: Include and exclude filter are mislabeled on p:directory-list
The comment says `<!-- RegularExpression -->`. It should be "List of RegularExpression" or "RegularExpression*s*" or, perhaps, removed entirely.

### LINKED ISSUE #173: The attribute type on set-attributes is wrong
We declare it as a map of (QName,untypedAtomic). That's the wrong way around, it should be (QName,anyAtomicType).

ndw:
Fix #110, fix #166, fix #173.


xml-project:
I think the change to step-json is obsolete because we moved p:json-join to main step library. As we have decided to have no other JSON step, this document could be deleted.
Will raise another issue for this.

Issue #159, opened 2019-07-05
Treatment of inputs to p:xslt


xml-project:
As issue #61 is heavily concentrated on what comes **out** of a p:xslt, I opened up a new issue about what's going **in**. The following questions result of my attempt to implement p:xslt for XSLT 3.0 and should be settled.

1. As the "Global Context Item" and the "Initial Match Selection" are now two different concepts in the case of the "Apply templates invocation": Do we want to expose this distinction to the interface of p:xslt? Or do we say: All document of port `source` form the initial match selection and the first one is the global context item?

2. In the case of the "Call template selection" the global context item is used as the context item for the named template. If we use the first document of source port as the global context item in this case: Is it an error if there are more documents on this port? Or do we use the "default collection strategy" of XProc 1.0 for this case? Or do we simply ignore them?

 3. In XProc 1.0 (and XSLT 2.0) the presence of a template name was the indicator for calling a template with this name. In XSLT 3.0 the call template selection does not need a template name anymore because there is a default template name defined. Do we kept the logic "template name present -> call template selection" (so the default template name has to be written) or do we use an additional option to switch between the two modes?

4. In XProc 1.0 we had the rule that the sequence of documents on source port is the default collection. I used that a lot (with XSLT 2.0) but as the sequence is also the Initial Match Selection I am not sure it will work out as the XSLT processor will apply the templates to **all** items in the initial match selection. Maybe we could use on option select the desired behaviour?

I do not claim, that this are all the question we have to settle for inputs on XSLT 3.0.

xml-project:
I think I cover all points with PR #178, so I can close this.
Feel free to reopen this or open a new issue.

Issue #61, opened 2019-04-18
Open questions on p:xslt


xml-project:
Unless I missed something important, some questions about p:xslt seems to be open:

1. An XSLT 3.0 processor is invoked with an `initial match selection` which can be a sequence of XDM items. In the current signature of p:xslt port `source` is sequence="false". Do we stay we the item only approach or do we allow a sequence of documents to appear for port `source`? [For correction please see below]

2. An XSLT 3.0 processor may return a sequence of XDM items as primary results, but port `result` is defined as sequence="false". Do we want to change this?

3. As far as I understand our discussion yet, we still want to support XSLT 1.0 and 2.0 processors with p:xslt. If so, I think it should be a dynamic error, if the invoked processor is not 3.0 and there is not exactly one document on port `source` and this has to have an XML content-type.

Conal-Tuohy:
Aren't the multiple output documents of an XSLT 3.0 transformation always _secondary_ results?

If so, these multiple items are not a sequence of documents, but a single document containing a sequence, and it seems to me the interpretation of the sequence should be dealt with according to the serialization specification. https://www.w3.org/TR/xslt-xquery-serialization-31/#serdm

xml-project:
> Aren't the multiple output documents of an XSLT 3.0 transformation always secondary results?

Not as far as I understand the XSLT 3.0 specs (and the Saxon 9.8.xx) interface. It [reads](https://www.w3.org/TR/xslt-30/#invoking-initial-mode):

> and the results of processing each item are then concatenated into a single sequence, respecting the order of items in the input sequence.

The creation of a secondary result with xsl:result-document is handled in a completely different section, so they are independent to my understanding.
Did I miss something?

Conal-Tuohy:
I was talking about the _outputs_ of the XSLT transformation, rather than the input.

As I understand it, the only way an XSLT 3 transform can produce a sequence of _documents_, is by using `xsl:result-document`; if it produces a sequence as its "principal result", then that should be output as a single document, and should not therefore be an XProc sequence, so `sequence='false'` is correct for the `result` port of `p:xslt`. That was the point I was trying to make.

xml-project:
I think the quoted sessions talks about _output_ or what other meaning could I attach to "results of processing"? (BTW I was using 'document' in the XProc sense = An XML document, a TEXT document, a JSON document etc.)

Could you please cite any passage from the XSLT 3.0 spec to support your understanding?

I tried the following with Saxon 9.8.x:
[...]

PR #178, opened 2019-07-28
Rewriting p:xslt for issues 61 and 159

xml-project:
I rewrote p:xslt to fix the problems in #61 and #159 . 
I introduces three subsections, so invocation of XSLT 1.0, 2.0 and 3.0 is now treated separately. As the terminology from 2.0 to 3.0 changed dramatically, this approach makes it easier to use the relevant terminology without interfering with the others.
Proposal: If you do not think it is completely non-sense, I would suggest to approve this PR and then file issue, so we do lot loss track of the necessary changed.


gimsieke:
Maybe attach a role="editorial" to the final note before merging.

PR #185, opened 2019-08-03
Fixed error codes in p:xslt

xml-project:
Fixed wrong error codes according to wiki error list.


xml-project:
I do not think this is controversial, is it? Could somebody please have a look at it.

Issue #187, opened 2019-08-04
Missing ruling in p:xslt


xml-project:
In the [specs](http://spec.xproc.org/master/head/steps/#c.xslt) we say:

> Option output-base-uri sets the base output URI per XSLT 3.0 specification. If a final result tree is constructed, this URI is used to resolve a relative URI reference. If no value is supplied for output-base-uri, the base URI of the first document in the source port's sequence is used.

This is taken verbatim from the 1.0 specs, but: What happens if there is no document on port `source` because  it is bound to `p:empty`?

If I remember right, XMLCalabash and MorganaXProc found two different solutions to this problem.

What to do? We could use

1. base-url of the stylesheet
2. base-uri of p:xslt
3. base-url of p:declare-step
4. base-uri of pipeline.

Suggestions?

ndw:
I think the XSLT answer is that it's the base URI of the stylesheet.

xml-project:
Will add a note to the specs.

xml-project:
Fixed with #191

PR #190, opened 2019-08-05
Attempt to fix 186 and 187

### LINKED ISSUE #186: XC0056 used for two different errors
As in XProc 1.0 we currently use XC0056 to mark that an initial mode OR an initial template name can not applied to a stylesheet.
I think this should be two different error so they can be caught by different p:catch.

xml-project:
Attempt to fix #186 and #187


ndw:
There's a typo in the first paragraph, `C008`.

PR #191, opened 2019-08-05
Attempt to fix 186 and 187

### LINKED ISSUE #186: XC0056 used for two different errors

xml-project:
Attempt to fix #186 and #187

### DIFF
--- a/steps/src/main/xml/steps/xslt.xml
+++ b/steps/src/main/xml/steps/xslt.xml
@@ -49,7 +49,7 @@
     <para>If no value is supplied for <option>template-name</option> option an “Apply-template
       invocation” is performed. The documents that appear on <port>source</port> are taken to be the
       initial match selection. If a value is supplied for the <option>initial-mode</option> option,
-      this value is used as the initial-mode for the invocation. <error code="C0056">It is a
+      this value is used as the initial-mode for the invocation. <error code="C008">It is a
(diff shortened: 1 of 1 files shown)


steps/src/main/xml/steps/xslt.xml
```diff
@@ -94,8 +96,10 @@
         code="C0007">It is a <glossterm>dynamic error</glossterm> if any key in
         <option>parameters</option> is associated to a value with is not an instance of the XQuery
         1.0 and XPath 2.0 Data Model, e.g. with a map, an array, or a function.</error>
-      <error code="C0056">It is a <glossterm>dynamic error</glossterm> if the specified initial mode
-        or named template cannot be applied to the specified stylesheet.</error>
+      <error code="C0008">It is a <glossterm>dynamic error</glossterm> if the specified initial mode
```
gimsieke: What I’m going to say is maybe a bit of nitpicking, and the wording has been like this before:

But do you really apply a mode or a template to a stylesheet?

Isn’t it more like “the stylesheet doesn’t support a given mode” or that ”the stylesheet doesn’t provide a given template”, respectively?

xml-project: Created issue #192 for this.

PR #198, opened 2019-08-27
Attempt to fix 197

### LINKED ISSUE #197: Add sort key option to p:text-sort
@xml-project pointed out that the `stable` option on `p:text-sort` does not makes sense as it is specified. Lines are either the same or not.

What we can do is add an option that defines a sort key as an XPath expression on the step. Proposal: `select`. This expression would act on every line (so you can do things like `substring(., 1, 3)` to sort on the first three characters). I think this is very useful and `stable` will again make sense.

xml-project:
Attempt to fix #197 

- I used "sort-key" because it feels more natural to me then "select". The only "select" in the step library I found is on p:filter and there it really select a portion of a document. If I am the only one arguing against "select" I am perfectly prepared to give in.
- My proposal keeps option `data-type` but removes the default value, so there a two ways to enforce a special sort mechanism:
````
<p:sort sort-key="xs:integer(substring(., 1, 5))" />
````
or
````
<p:sort sort-key="substring(., 1, 5)" data-type="numeric" />
````
As far as I can see, data-type on xsl:sort is only for legacy reason, so if you opt to remove `data-type`, I am perfectly happy with this.

### DIFF
--- a/steps/src/main/xml/steps/text-sort.xml
+++ b/steps/src/main/xml/steps/text-sort.xml
@@ -13,7 +13,6 @@
     <p:option name="order" as="xs:string" select="'ascending'" values="('ascending', 'descending')"/>
     <p:option name="case-order" as="xs:string?" values="('upper-first', 'lower-first')"/>
     <p:option name="lang" as="xs:language?"/>
-    <p:option name="data-type" as="xs:string?" values="('text', 'number')"/>
     <p:option name="collation" as="xs:string"
       select="'https://www.w3.org/2005/xpath-functions/collation/codepoint'"/>
     <p:option name="stable" as="xs:boolean" select="true()"/>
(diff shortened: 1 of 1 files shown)


xml-project:
@ndw Any objections especially concerning implementability?

xml-project:
@ndw I thought about this, but the [XSLT 3.0 specs](https://www.w3.org/TR/xslt-30/#comparing-sort-keys) says:

> If the data-type attribute has any other effective value, then this value must be an EQName denoting an expanded QName with a non-absent namespace, and **the effect of the attribute is implementation-defined.**

So we need to spell out ourselves the behaviour we prefer. I am not sure, it is worth it. Any suggestions?

xml-project:
Since I do not know, how to go on, let's do a little survey:

(A) Change 'data-type' as in XSLT 3.0 and spell out, which behaviour is expected.

(B) Leave 'data-type' as it is in the PR.

(c) Remove 'data-type' completely.

Please take a short moment and give me a letter.
Thanks.

gimsieke:
Slight preference for (A), with support for the `http://www.w3.org/2001/XMLSchema` namespace URI (others implementation-defined). If it’s not too expensive implementation-wise.

ndw:
Per 29 Aug 2019 editor's call: remove the 'data-type' attribute.

Issue #196, opened 2019-08-07
Inconsistency in the validation document


xml-project:
In the preamble it reads:

> the mandatory report document should be valid with respect to the [XVRL] schema.

But for p:validate-with-schematron it reads:

> In addition to the mandatory XVRL report, a Schematron Validation Report Language (SVRL) report should be provided on the report port.

This seems to be the reason why @gimsieke and me disagree in #195

gimsieke:
I don’t see this contradiction. 

There is a mandatory report document that **should** be valid XVRL (I didn’t write **must** because I thought that processors shouldn’t be considered non-conformant if they sloppily generate not-fully-valid XVRL). But at least they should try their best in producing XVRL vocabulary in the primary, mandatory report document.

Then there may be additional report documents in other vocabularies. The Schematron validation should also emit an SVRL document on the `report` port in addition to the mandatory XVRL.

I’m pretty sure there is some misunderstanding among us but I don’t see yet how you interpret the text or what I fail to recognize as contradictory.

xml-project:
Obviously we have to completely different readings of this: I read "should be valid with respect ..." as "might be XVRL, but also might be something different".

But if there is a "mandatory XVRL report" I would expect this step to produce valid "XVRL".

If you don't think its contradictory, feel free to close this issue or change the text, so stupid people like me are not bothered.

I came to the conclusion that this whole discussion is simply void as we talk about an optional step. If think user will not want it, I am free not to implement it. So: Go on.

gimsieke:
I propose then to rephrase it:

> Unless another serialization format is requested, the mandatory report document **must** be an [XVRL] document.

So we don’t say anything about validity here.

I wouldn’t rule out the possibility of JSON or turtle text output for the primary report (at user’s request, and subject to availability of these other XVRL serializations). But if we want to do without these hypotheticals in this version of the spec, we can simply say:

> This primary report **must** be an [XVRL] document.

mricaud:
Hi folks ! 
Maybe toi late but IMHO, it would ne great that any validation step has a mandatory ans valid XVRL output port. This allow users to merge multiple XVRL into a single one if they want the whole validation report at once.
Off course this doesn't prévention to have other kind of report like SVRL for schematron, but as a specific output port si people that don't care about XVRL van access other specific report.
Maybe I missed some considerations but that was my feeling :)

gimsieke:
[...]

Issue #195, opened 2019-08-06
I don't think we should call out skeleton implementation in p:validate-with-schematron 


xml-project:
I do not see any reason why we state:

> An XProc processor that implements this step should at least support the [Schematron Skeleton] implementation.

What is special that we call out this implementation here, as the only implementation we mention in the whole spec?

I think we should remove the whole paragraph.

ndw:
Fine by me.

gimsieke:
It is important that there is an implementation that can include and execute XSLT. This behavior can be triggered by setting allow-foreign to true in the set of compile parameters. There are other params that users have come to rely on and that won't probably be supported by any random XPath-only implementation. I will respond later tonight in greater detail since I'm mobile-only atm.

gimsieke:
Suggestion:

> There are multiple Schematron implementations. An XProc processor that implements this step **must** be able to use an implementation that generates XSLT code and that offers the `xslt2` query binding. In addition, this Schematron implementation **should** be able to pass non-Schematron markup, including XSLT code, to the compiled XSLT stylesheet verbatim. The Schematron implementation **should** also be able to include machine-readable, namespace-aware XPath location expressions in its reports.

We can move the Skeleton citation to here: 

> The parameters map may contain two special entries, `c:compile` and `c:xvrl`, both are maps. If a code-generating implementation such as [[Schematron Skeleton](http://spec.xproc.org/master/head/validation/#schematron-skeleton)] is used, the entries of the `c:compile` map, for example `allow-foreign`, will be passed to the code generator.

xml-project:
@gimsieke I see your point and that you want all this features. But frankly: I think this should be a quality of implementation and we should not raise the bar for implementations beyond standard compatibility. Think of somebody coming up and demanding that any implementation of `p:xslt` **must** be streaming. Hopefully we would not accept it.
So still: I propose to delete the whole paragraph and not demanding anything beyond the respective standard.

dmj:
I don't understand why there are any requirements **in the specification** that go beyond simple conformance as defined in ISO 2016 at all. It is up to the implementors to provide an implementation that suits their customers and/or users, isn't it?

gimsieke:
The Schematron spec is fuzzy in some regards. 

[...]

PR #213, opened 2019-09-12
Update Schematron

### LINKED ISSUE #194: Signature of p:validate-with-schematron is wrong
Port `result` of p:[validate-with-schematron](http://spec.xproc.org/master/head/validation/#c.validate-with-schematron) is defined as `application/xml` while the prose says:

> The result output from this step is a copy of the input.

Therefor I would expected, that `source` and `result` would have the same signature.
What did I miss?

gimsieke:
Fix #194 
Address #195, #196


gimsieke:
> 
> 
> I think there are some issues with this PR:
> 
>     1. The note on "p:validate" is removed, but the introduction still mentions this step.

Ok, removed it. I don’t know why my recent push doesn’t show up here though.

> 
>     2. While we currently agreed that XVRL **should** be used, the PR says it **must** be used. I think I can not agree on this change as long as I do not know, what XVRL is. (This spec on this is still a blind text). I think we should stay with "should" for a while.

I put “should” in.

> 
>     3. We agreed that p:validate-with-schematron can have a SVRL report and this is mentioned in the prose, but the option on p:validate-with-schema for this is missing.

We need to discuss whether a separate `svrl` port on `p:validate-with-schematron` would be better? Otherwise I’d be inclined to provide it a key of the `c:xvrl` map (see also remark below) in which the pipeline author can specify whether they want SVRL in addition to or instead of XVRL on this port.

Regarding the recent discussion about nested maps and how they are cumbersome to produce for pipeline authors: Should we provide distinct map options on `p:validate-with-schematron` (or all validation steps) for the compilation and the XVRL generation parameters?

> 
>     4. As an implementer I do not understand what the paragraph starting with "If the Schematron implementation produces SVRL natively..." because (1) I do not know what 'natively' means here and (2) what is meant by 'influencing the conversion'. Do we really need this para?

“Natively” should say: The Schematron implementation produces an SVRL report, like most Schematron implementations do, not an XVRL report by default. Some mappings from SVRL to XVRL may be customized, such as the mapping from `@role` or other Schematron/SVRL attributes to the XVRL error vocabulary (see https://github.com/xproc/xvrl/issues/2). There should be standardized map keys for the parameter map that influences XVRL generation, for ex. `map-to-severity` for the SVRL attributes that should be mapped to XVRL severity.

gimsieke:
#194 and #257 have been auto-closed, #195 and #196 are still open

Issue #114, opened 2019-06-11
Add p:compress/p:uncompress steps


ndw:
Compress takes serialization parameters and compresses the serialized representation using [insert favorite compression flavor here]. Uncompress reverses the process. Each takes exactly one document.

eriksiegel:
Is there any compression format that we define as mandatory here?
Or are all the supported compression formats completely implementation defined?

gimsieke:
I’d support making zip (we already have it for archives, we can also use it on single files) and gzip mandatory.

eriksiegel:
Then I suggest to _only_ prescribe supporting `zip ` because we do this in p:archive etc. also.

eriksiegel:
But zip is an archive format (compressing *and* container format for multiple files). Excuses for my ignorance,but is there also something like a zip *compression* format and if we state compression format zip is that clear enough for implementors?

gimsieke:
I don’t think there is an independent zip compression format. It can use a variety of compression methods but I think it always creates an archive (that may contain a single file).

We can explain that using `zip` as a compression method is equivalent to creating a zip archive with a single entry.

Because of its widespread use, I’d still support also stipulating that an implementation provide the `gzip` format.

eriksiegel:
> Because of its widespread use, I’d still stipulate that an implementation support the gzip format.

Would you like to add this to the mandatory formats for `p:archive` and `p:unarchive` also? 

I would like to hear the opnion of our implementors @ndw  and @xml-project also about this.

gimsieke:
No, `gzip` doesn’t do archiving. If we wanted to cover the archive format that often comes gzipped, we’d need to require `tar` support in the archive steps. I’m not against requiring tar over there.

eriksiegel:
Let's leave the `p:(un)archive` steps as they ar now for the moment.

Then I suggest specifying two mandatory formats (beside any implementation defined ones) for `p:(un)compress`:
* `zip` which results in a zip file with a single entry. `p:uncompress` expects a zip file with a single entry, multiple entries result in an error.
* `gzip` that compresses usign the gzip format

@ndw, @xml-project ?

gimsieke:
Let’s discuss whether `p:uncompress` should behave as `p:unarchive` when the compression method is `zip`. 

I think this satisfies user expectations. 

If we think this is weird then we should reconsider requiring `zip` as a `p:compress` method.

eriksiegel:
[...]

PR #214, opened 2019-09-13
Issue 114 compress

eriksiegel:
Attempt to describe the `p:compress` for #114. If this is ok I'll create a proposal for `p:uncompress` as well.

Please niotice what I wrote about the preservation of the document properties. This seemed to me the logical thing to do here but maybe we don't agree on that?


steps/src/main/xml/steps/compress.xml
```diff
@@ -0,0 +1,67 @@
...
+
+  <p:declare-step type="p:compress">
+    <p:input port="source" primary="true" content-types="any" sequence="false"/>
+    <p:output port="result" primary="true" content-types="any" sequence="false"/>
+    <p:option name="format" as="xs:QName" select="'gzip'"/>
+    <p:option name="serialization" as="map(xs:QName,item()*)?"/>
+    <p:option name="parameters" as="map(xs:Qname, item()*)?"/>
+  </p:declare-step>
+
+  <para>The <code>p:compress</code> step first serializes the document appearing on its
```
gimsieke: Please have a look how step names are tagged at other places in the spec. I think it’s `<tag>p:compress</tag>` but I’m not quite sure.

steps/src/main/xml/steps/compress.xml
```diff
@@ -0,0 +1,67 @@
...
+  <title>p:compress</title>
+
+  <para>The <code>p:compress</code> step serializes the document appearing on its
+      <code>source</code> port and outputs a compressed version of this on its <port>result</port>
+    port.</para>
+
+  <p:declare-step type="p:compress">
+    <p:input port="source" primary="true" content-types="any" sequence="false"/>
+    <p:output port="result" primary="true" content-types="any" sequence="false"/>
+    <p:option name="format" as="xs:QName" select="'gzip'"/>
```
gimsieke: Don’t have the time right now to check – are we using strings as default values for `xs:QName` in other places? I mean, we have the rule about treating uncolonized strings as QNames, but can we use it sloppily for default values? Or do we need to write `select="xs:QName('gzip')"` for the default value?

steps/src/main/xml/steps/compress.xml
```diff
@@ -0,0 +1,67 @@
...
+    <p:option name="serialization" as="map(xs:QName,item()*)?"/>
+    <p:option name="parameters" as="map(xs:Qname, item()*)?"/>
+  </p:declare-step>
+
+  <para>The <code>p:compress</code> step first serializes the document appearing on its
+      <code>source</code>. It then compresses the outcome of this serialization and outputs the
```
gimsieke: `<port>source</port> port` (that is, other tag and add the word „port“). Also please use `<port>` and `<tag>` (or whatever is canonical) below

steps/src/main/xml/steps/compress.xml
```diff
@@ -0,0 +1,67 @@
...
+      </listitem>
+    </varlistentry>
+    <varlistentry>
+      <term><option>serialization</option></term>
+      <listitem>
+        <para>The <option>serialization</option> option is provided to control the serialization of
+          content before compression takes place. If the document to be stored has a
+            <code>serialization</code> property, the serialization is controlled by the merger of
+          the two maps where the entries in the <code>serialization</code> property take precedence.
+          Serialization is described in <biblioref linkend="xproc30"/>.</para>
```
gimsieke: Can we link to a specific section of the XProc 3.0 spec? I think it’s quite rude if we “arbitrarily” split the spec up into multiple parts and thereby lose the ability to link directly to sections of the main spec.

steps/src/main/xml/steps/compress.xml
eriksiegel: > are we using strings as default values for xs:QName in other places

Yes, absolutely. You probably didn't notice (?), but most of the text/spec for p:compress was copy/pasted and adapted.

steps/src/main/xml/steps/compress.xml
eriksiegel: A quick search shows that both `<code>` and `<tag>` are used. So its not consistent. But I changed it to `<tag>`.

steps/src/main/xml/steps/compress.xml
eriksiegel: DOne.

steps/src/main/xml/steps/compress.xml
eriksiegel: No, we can't. This is a biblioref and it links to the full resource.

Anyway, this was also copy/paste text and it happens in more places. So if you think this is important, maybe we should open an issue with "finishing touches" or something like that and gather all editorial things that we need/want to do before finally publishing?

eriksiegel:
@ndw @xml-project @gimsieke Could we please either discuss something about it or approve this PR?

xml-project:
@eriksiegel Sorry Erik, I was not aware of this PR. My interface says you ask only @gimsieke for approval.
Is there a way to look at your final PR? If so, I will have a look at it tomorrow afternoon.

eriksiegel:
@xml-project Weird, spec authors is also listed as reviewer, and that's us, right? What goes wrong here?

As always I have no github pages available, but its a single file describing `p:compress` so that shouldn't be too hard.

PR #219, opened 2019-09-21
Attempt to fix issue 217

### LINKED ISSUE #217: Should/Does p:store with "file://" also create folders?
From a private conversion:

In `p:store` we just say that it is an error, if the resource can not be stored (which is obvious), but we do not say, what the step should undertake to make storing possible.

At least for "file:///" the question might come up, whether a processor will create all the folder necessary for storing the resource with the given URI.

I do not argue, a processor should do this, but I argue that we should **say** what a processor is expected to do or not to do.

xml-project:
Attempt to fix #217 : If "file:" is supported, non existing folders should be created.

### DIFF
--- a/steps/src/main/xml/steps/store.xml
+++ b/steps/src/main/xml/steps/store.xml
@@ -22,7 +22,9 @@ of a syntactic shortcut
 value).</para>
 
 <para>The step attempts to store the document to the specified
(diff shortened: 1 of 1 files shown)


xml-project:
But there were typographic quotes :-))

PR #221, opened 2019-09-23
Attempt to resolve issue #114/uncompress

eriksiegel:
See #114


eriksiegel:
@xml-project You're right about the missing error condition. But this applies to p:compress as well. I will try to amend this in a further PR.

PR #224, opened 2019-09-29
Attempt to fix 220

### LINKED ISSUE #220: Default value for @merge on p:set-properties
In the last days I wrote a bigger pipeline, where I had to modify document-properties a lot, e.g. attaching a base-url or adding serialization properties. These all had to be done with `merge=true()`, because the default value is false.

I think we should reverse this decision and make 'true' the default value.

xml-project:
Attempt to fix #220 
- Option merge is now true by default
- Little change of prose to improve clarity.

### DIFF
--- a/steps/src/main/xml/steps/set-properties.xml
+++ b/steps/src/main/xml/steps/set-properties.xml
@@ -12,7 +12,7 @@ properties on the source document.</para>
    <p:input port="source" content-types="any"/>
    <p:output port="result" content-types="any"/>
    <p:option name="properties" required="true" as="map(xs:QName,item()*)"/>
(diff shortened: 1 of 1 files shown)


steps/src/main/xml/steps/set-properties.xml
```diff
@@ -23,7 +23,8 @@ adjusted property values.</para>
 
 <para feature="p-set-properties-merge">If the <option>merge</option>
 option is true, then the supplied properties are added to the existing
-properties. If it is false, the document’s properties are replaced by
+properties overwriting already existing values for a given key.
```
gimsieke: maybe a comma before “overwriting”

PR #230, opened 2019-10-03
Attempt to fix 880 (main specs)

xml-project:
Attempt to fix 880 (main specs). Please mind there is a corresponding pr there.


steps/src/main/xml/steps/set-properties.xml
```diff
@@ -26,6 +26,11 @@ option is true, then the supplied properties are added to the existing
 properties. If it is false, the document’s properties are replaced by
 the new set.</para>
 
+<para><error code="D0070">It is a <glossterm>dynamic error</glossterm> if a value is 
+assigned to the <code>serialization</code> document property can not be converted
```
gimsieke: … document property can not be converted …
→
… document property that cannot be converted …

Otherwise approve

Issue #239, opened 2019-10-16
Provide defaults for manifest and archive ports of p:archive?


eriksiegel:
The way the `p:archive` step is defined now means you always have to provide connections for the `manifest` and `archive` input ports, even when you don't need them (which happens, quite often I think, especially the `archive` port). So you'll have to bind them to `p:empty` now.

Shall we therefore provide these ports with a *default* `p:empty` connection?

One problem (probably) (this one is for @ndw): Can the spec formatting pipeline handle default connections on input ports in a step specification?

ndw:
My first reaction to this is pretty negative. I think it's going to be very confusing. If we provide a default binding, it will only apply if there isn't a default binding. So lots of the time, you'll get whatever happens to be on the DRP even if we specify a default binding. When that isn't appropriate, you'll still have to make the binding explicit.

Or am I misunderstanding something?

eriksiegel:
I don't quite get you I'm afraid, I do not understand what this has to do with the DRP. 

Here is what _I_ mean: When you invoke `p:archive` you now **always** have to provide bindings for the `archive` and `manifest` port, even when you don;t need them, which will be quite often. So you **always** have to write:

```xml
<p:archive>
  <p:with-input port="archive">
   <p:empty/>
  </p:with-input>
 <p:with-input port="manifest">
   <p:empty/>
  </p:with-input>
</p:archive>
```

Especially for the `archive` port, an empty connection will very often be the case. So you have to write the `p:with-input port="archive` part over and over again. Not a world class problem but we have means to make life simpler, don't we?

So why not just _define_ the archive` and `manifest` ports of `p:archive` with a default `p:empty` connection?

xml-project:
@ndw wrote:

> The default bindings for archive and manifest would not apply here. Because the XSLT step provides a default readable port, that's the input on both of those ports.

Would you mind to explain this please, because I do not get it. Ports `archive` and `manifest` are non-primary port, so port `result` of p:xslt to my reading **DOES NOT** provide the DRP for the ports on manifest. Where is my mistake?

gimsieke:
Can't we say that unconnected non-primary ports will always connect to p:empty if no explicit or default connection is given? (If sequence=false, p:empty will cause a dynamic error, but this is Ok).

xml-project:
Don‘t you think that is to wide a solution because it would apply to every user-defined step ever written. I would tend to Eriks suggestion to do this step wise.
[...]

PR #242, opened 2019-10-17
#239 provide defaults for non-primary input ports

eriksiegel:
Defaults for the non-primary input ports of `p:archive`. See #239


ndw:
I don't know what's wrong on Travis; I'll investigate

ndw:
It was a transient problem at Travis.

Issue #253, opened 2019-11-04
Some file steps lack proper error codes and other editorial stuff


eriksiegel:
There are some file steps (I just spotted them there, maybe there are others?) that have not properly filled error codes. My omission, I'll amend it.

eriksiegel:
There seemed to be some other smaller editorial stuff to do: Intro to the error section was not ok, typos, an editorial note that can go.

PR #254, opened 2019-11-04
Attempt to resolve 253

eriksiegel:
Editorial update on the file steps. #253

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -347,7 +347,7 @@ steps is assumed; for background details, see
             <option>href</option> option does not exist, cannot be accessed or is not a file or directory.</error></para>
     </listitem>
     <listitem>
-      <para><error code="XXXXX">It is a <glossterm>dynamic error</glossterm> if an attempt is made to delete a non-empty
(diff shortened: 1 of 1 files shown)


eriksiegel:
@gimsieke No I wasn't aware of that list, thanks 👍 . Since it is manually maintained, is it up to date do you think?

I'll change the error codes to new ones, no problem.

eriksiegel:
@gimsieke, would you mind to review again?

gimsieke:
Until [Aug 5](https://github.com/xproc/3.0-steps/pull/185#pullrequestreview-270584897) I wasn’t aware of the list either…
Before that, Norm and Achim have maintained in secret for about two years!

xml-project:
> Before that, Norm and Achim have maintained in secret for about two years!

No we have not. It would be a very stupid thing to keep a list secret _**in a public wiki**_, wouldn't it?

gimsieke:
> 
> 
> > Before that, Norm and Achim have maintained in secret for about two years!
> 
> No we have not. It would be a very stupid thing to keep a list secret _**in a public wiki**_, wouldn't it?

If you have co-editors like Erik and me, apparently you can!

Issue #203, opened 2019-08-28
Missing case in p:archive?


xml-project:
Unless I overlooked something, there seems to be a missing case. The [specs](http://spec.xproc.org/master/head/steps/#cv.archive-zips) say on 'update':

> - Any entry in the manifest that does not exist in the ZIP file provided on the archive port is added.
> 
> - For any entry that does exist in the ZIP file provided on the archive port, a timestamp compare is done. When the entry in the manifest is newer than the entry in the ZIP file provided on the archive port, it is replaced. Documents that are either specified in the manifest or come from the source port have no timestamp and will always cause a replace to be done.
> 

Now image I run update on a zip containing entries, that are not mentioned in the (constructed) manifest? What do we say about them?

I suppose these entries should be copied to the archive appearing on port `result`, but should be moved to the end of the zip because the manifest imposes an order on the entries.

Right?

eriksiegel:
Sounds ok to me. Bit of an edge case I suppose.

xml-project:
@ndw Yes, order makes it complicated and results in a lot I/O. But the specs say:

> The p:archive step **should** strive to retain the order of the c:entry elements when constructing the archive. For instance, an e-book in EPUB format has a non-compressed entry that must be first in the archive. It should be possible to construct such an archive using p:archive.

(Hint: The should in the last sentence is not in terminology).

I pretty sure creating ePub is a requirement for some users.

eriksiegel:
Decision: Manifest imposes order, everything else is implementation defined
Manifest entries come first

eriksiegel:
Work moved to #251

PR #255, opened 2019-11-06
Attempt to resolve issue 251

eriksiegel:
This turned out to be quite a change due to our decisions that base URIs must be absolute. I had to change `p:archive`, `p:archive-manifest` and `p:unarchive`.

All three now have a `relative-to` option that can be used to set the base URI for the files *in* the archive.

I also made the choice to have all relevant URIs to be normalized first by `p:urify()`. This applies to values passed to the `relative-to` option and the values of the manifest's `c:entry/@href` attributes. IMHO this is in line with the decisions we made about this for setting base URIs on our last call.

### DIFF
--- a/steps/src/main/xml/references.xml
+++ b/steps/src/main/xml/references.xml
@@ -20,6 +20,7 @@
     <bibliomixed xml:id="rfc2119"/>
     <bibliomixed xml:id="rfc2616"/>
     <bibliomixed xml:id="rfc2617"/>
+    <bibliomixed xml:id="rfc3986"/>
     <bibliomixed xml:id="unicodetr17"/>
     <bibliomixed xml:id="tidy"/>
     <bibliomixed xml:id="tagsoup"/>
--- a/steps/src/main/xml/steps/archive-manifest.xml
+++ b/steps/src/main/xml/steps/archive-manifest.xml
@@ -51,19 +51,19 @@
       whose value is not valid for that key.</error></para>
 
   <para>The <option>relative-to</option> option, when present, is used in creating the value of the
-    manifest's <code>c:entry/@href</code> attribute.</para>
-  <para>The value provided for the <option>relative-to</option> option will before use always be
-    normalized using the <function>p:urify()</function> function. After this normalization its value
-      <rfc2119>must</rfc2119> be an absolute URI. </para>
-  <para></para>
+    manifest's <code>c:entry/@href</code> attribute. <error code="D0064">It is a <glossterm>dynamic
+        error</glossterm> if the option is not a valid URI according to <biblioref linkend="rfc3986"
+      />.</error> If the option is relative, it is made absolute against the base URI of the element
+    on which it is specified (<tag>p:with-option</tag> or the step in case of a syntactic shortcut
+    value).</para>
 
   <para>The generated manifest has the format as described in <xref linkend="cv.archive-manifest"/>.
     Implementations <rfc2119>must</rfc2119> supply an <tag>c:entry</tag> element and its
       <code>name</code> attribute for every entry in the archive. The value of the generated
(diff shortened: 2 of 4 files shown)


eriksiegel:
>In https://github.com/xproc/3.0-specification/pull/896/files, I already used D0074ff.

Well, since a week I'm using the error list https://github.com/xproc/3.0-specification/wiki/Error-Codes to get a number for a new error. And this number seemed to be unused... So I suppose you forgot to mention it there?

>Don’t the steps throw C00… errors anyway?

The steps now already specify raising D numbers when it is a common error. And specifying a non absolute base URI seemed a common error to me given the decisions made on last call. Both to be raised by the core *and* the steps. 

>And if p:urify() itself already throws errors if the URI is not absolute and compliant with RFC 3986, do the individual steps have to throw errors?

I didn't realize `p:urify()` raised errors, so you're probably right about steps not having to do that themselves. I'll revisit my changes and see how this affects the descriptions.

gimsieke:
> 
> 
> > In https://github.com/xproc/3.0-specification/pull/896/files, I already used D0074ff.
> 
> Well, since a week I'm using the error list https://github.com/xproc/3.0-specification/wiki/Error-Codes to get a number for a new error. And this number seemed to be unused... So I suppose you forgot to mention it there?

I was aware that this is a race condition that may lead to conflicts, but I preferred not to add error messages to the list before the corresponding pull requests have been accepted.

> 
> > Don’t the steps throw C00… errors anyway?
> 
> The steps now already specify raising D numbers when it is a common error. And specifying a non absolute base URI seemed a common error to me given the decisions made on last call. Both to be raised by the core _and_ the steps.

If you say so. I don’t know the system for error prefixes tbh. We need a metaspec for the prefix/numbering scheme ;)

> 
> > And if p:urify() itself already throws errors if the URI is not absolute and compliant with RFC 3986, do the individual steps have to throw errors?
> 
> I didn't realize `p:urify()` raised errors, so you're probably right about steps not having to do that themselves. I'll revisit my changes and see how this affects the descriptions.

Maybe first settle what `p:urify()` does and which errors it throws.

eriksiegel:
@gimsieke : I did the following:

* I removed the specific error codes in `p:archive`, `p:archive-manifest` and `p:unarchive` for things that are processed by `p:urify()`. 
* I removed D0074 from the error code wiki page
* There seem to be more/other situations that need error codes, I opened an issue for this to discuss: #256 
* I will put the error code maintenance/policy/claiming/... on the proposed Cologne agenda

My suggestion is to accept the PR with these changes and re-adjust if we find out we do need to say more about specific errors.

gimsieke:
I agree that we should discuss it. 

A general rule may be to run some normalization and/or compliance checking function on all `xs:anyURI` options and on the base URIs of all documents.

I’d be hesitant to stipulate that we call `p:urify()` by default in these circumstances. One reason is that it will not be obvious to the processor whether it should be better called with no 2nd argument (which means that relative paths will be resolved against the current working directory if available) or with the context document’s base URI as 2nd argument, or with something else as 2nd argument. This is probably the pipeline author’s call to make.

We should nevertheless stipulate that the processor may (or must?) check that every `xs:anyURI` option and every `base-uri` document property is an absolute URI that complies with RFC 3986.

For `@xml:base`, there is already a rule that it must comply with RCF 3986 (by and large), but it need not be absolute. Since `@xml:base` can be set within an opaque step, for example in `p:xslt`, the processor can’t be bothered IMHO to always check `@xml:base` for compliance. Maybe only when it is set by `p:add-attribute`. I think we can safely defer a check until someone uses this URI for a document property or for an option value.

eriksiegel:
You can find the buld step spec at: https://eriksiegel.github.io/3.0-steps/issue-251/head/steps/index.html

Look at the `p:archive`, 'p:archive-manifest' and `p:unarchive`

I included the changes for #256 & #203

PR #269, opened 2019-11-20
Attempt to fix 268

### LINKED ISSUE #268: Not sure whether p:unarchive is right
If I understand our consensus in Cologne regarding base URIs right, then a document flowing out of a step either has NO base URI (like c:result in p:count) or has a valid absolute URI according to the RFC specification. Right?

If so, we need to fix p:unarchive, because it is not certain, that every document coming out this step has an RFC valid URI. The base against which relative URI of the archives entry are resolved is valid (or an error is raised), but the entries name itself may raise the problem. Just think of an entry named "%gg ##". The resulting base URI of the document is not valid. Right?

As a solution I propose, that p:urify is implicitly called with the entrie's name as first param and either "relative-to" or the archive's URI as second param. It is an error, if the resulting string is not a valid URI.

We voted against implicit use of p:urify in Cologne but I think we need it in this case because the pipeline author has no way to make sure, the entries in the archive result in valid URIs.

Did I missunderstand or miss something?

xml-project:
Attempt to fix #268 @gimsieke please mind the hyphens ;-))

### DIFF
--- a/steps/src/main/xml/steps/unarchive.xml
+++ b/steps/src/main/xml/steps/unarchive.xml
@@ -92,9 +92,7 @@
     </listitem>
     <listitem>
       <para>The <option>relative-to</option> option, when present, is used in creating the base URI
-        of the unarchived documents. <error code="D0064">It is a <glossterm>dynamic
-            error</glossterm> if the option is not a valid URI according to <biblioref
-            linkend="rfc3986"/>.</error> If the option is relative, it is made absolute against the
+        of the unarchived documents. If the option is relative, it is made absolute against the
         base URI of the element on which it is specified (<tag>p:with-option</tag> or the step in
         case of a syntactic shortcut value).</para>
     </listitem>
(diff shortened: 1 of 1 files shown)


gimsieke:
I’m still not convinced that producing a URI like `file:///some/path/to/file.docx/word/document.xml` is the best choice for a base URI if `file:///some/path/to/file.docx` was the archive’s URI.

Analogy: When I use `<xsl:result-document href="relative.xml">` in `p:xslt`, the document on the secondary port will not have the base URI `file:///static/base/uri.xsl/relative.xml` (`file:///static/base/uri.xsl` being the base URI of the stylesheet) but `file:///static/base/relative.xml`, which may be used in principle for storing the result (if the user is allowed to create files in `file:///static/base/`). At least this URI is not _guaranteed_ to fail when attempting to store it to its base URI’s location.

An advantage of the proposed solution is that it is relatively easy to retrieve the original relative path when the archive URI is still known, for example using `substring-after(base-uri($archive-member), base-uri($archive) || '/')`.

We can discuss whether the `!/` separator that oXygen and Jar files use has some merit (`file:///some/path/to/file.docx!/word/document.xml`).

We can discuss it tomorrow in the call. I wouldn’t vehemently oppose the current proposal when there’s otherwise consensus.

Apart from this detail, I approve the PR.

I see it has already been merged while I was typing. We should still discuss it tomorrow.

PR #275, opened 2019-11-26
Attempt to fix 272

xml-project:


xml-project:
Hhm: Travis in Git says "build is in progress", but Travis itselfs says passed. Let's hope the last statement is right.

PR #279, opened 2019-11-29
Attempt to fix 274

### LINKED ISSUE #274: p:directory-list needs amendment
Trying to implement this (again), I found at least two passage, that need some more explanation.

- There is an attribute "uri" in c:file, c:directory and c:other which is not explained anywhere. 

- There is an attribute "content-type**s**" which is not explained. (I think it should be "content-type" anyway).

- Since "@uri" is mandatory, we should adapt "Example 1. Sample Directory List Output for a Single File"

xml-project:
Attempt to fix #274

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -236,6 +236,13 @@ steps is assumed; for background details, see
         <rfc2119>should</rfc2119> be provided by the implementation:</para>
       
       <variablelist>
+        <varlistentry><term><tag class="attribute">content-type</tag></term>
+          <listitem>
+            <para>The <tag role="attribute">content-type</tag> attribute contains
+              the content type of the respective file. The value “<literal>application/octet-stream</literal>”
+              is used, if the processor is not able to identify another content type.</para>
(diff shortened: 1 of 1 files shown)


step-file/src/main/xml/specification.xml
```diff
@@ -236,6 +236,13 @@ steps is assumed; for background details, see
         <rfc2119>should</rfc2119> be provided by the implementation:</para>
       
       <variablelist>
+        <varlistentry><term><tag class="attribute">content-type</tag></term>
+          <listitem>
+            <para>The <tag role="attribute">content-type</tag> attribute contains
+              the content type of the respective file. The value “<literal>application/octet-stream</literal>”
+              is used, if the processor is not able to identify another content type.</para>
```
gimsieke: Just a wording suggestion (native speakers may step in and provide a better suggestion):

> The value “application/octet-stream” will be used if the processor is not able to identify another content type.

(that is, “will be” and no comma)

Or:

If the processor is not able to identify another content type, the value “application/octet-stream” will be used.

PR #282, opened 2019-12-12
Clarify that for merge=false, “unsupplied” properties will be removed

### LINKED ISSUE #281: Does p:set-properties remove the base URI of a document?
@ndw and me disagree about the result of a test in the test suite. The original discussing is [here](https://github.com/xproc/3.0-test-suite/issues/327), but the question is this:

Suppose I have a `p:set-properties` with `merge=false` and the map in `properties` does not have an entry for `base-uri`. What about the base URI of the document on port `result`?

1. It does not have a base URI because there is no entry `base-uri` in `properties`. 

2. It does have the same as the document on port `source` because there is not entry for `base-uri` and therefor the base URI of the document is not changed.

I am not in favor of any answer, but would argue, that if (2) is the right answer, we should explicitly say in the specs, that the resulting document does not have a base URI in this case.

**Votes in please!**

gimsieke:
fix #281


xml-project:
I do think we should explictly say that the resulting document does not have a base URI in this case, because I expect this to be a cause for confusion.

gimsieke:
With “effectively” I wanted to shield myself from pedantic remarks claiming that the properties won’t be removed because a new document is created and it will be created without the properties in question.
But apparently the attempt wasn’t as “effective” as I had hoped. So if everyone if comfortable with the IMO slightly sloppy “will be removed from the document”, I’m fine with it, too.

ndw:
"Slightly sloppy". Zing! Them's fighting words. I now propose:

> <para feature="set-properties-preserves-some">If <option>merge</option> is true, document properties not overridden by settings in the <option>properties</option> map are preserved, otherwise the resulting document has only the <literal>content-type</literal> property and the properties specified in the <option>properties</option> map. In particular, if <option>merge</option> is false, the <literal>base-uri</literal> property will not be preserved.</para>

gimsieke:
…if merge is false _and if `base-uri` is omitted in the `properties` map_, the base-uri property will not be preserved.

gimsieke:
No, it’s ok. It doesn’t need to be omitted in order not to be preserved. I will use your words.

xml-project:
I am not sure that fits with the specs. We say:

> The value of the “base-uri” property identifies the base URI of the document;

so we should say, that the resulting document will not have a base URI.

gimsieke:
Acceptable now?

xml-project:
Thanks Gerrit!

Issue #16, opened 2018-04-22
p:cast-content-type needs an update


xml-project:
I think we should take a closer look at [p:cast-content-type](http://spec.xproc.org/master/head/steps/#c.cast-content-type), because it seems to be somewhat outdated by the development of XProc 3.0. Some of my thoughts:

1. In Prague we agreed that a text document is a text node surrounded by a document node. Since all encoding problems are solved when the text document is created, I think we no longer need a base64 encoding when casting a text document to an XML-document.
1. Since we define the loading of a JSON document via `fn:parse-json()`, it seems natural to me to use the inverse function `fn:serialize(., map{"method":"json"})` when a JSON document is casted to a text document.
1. XPath 3.1 defines `fn:json-to-xml()` which converts a JSON text to an XML-document. It is imho therefore obvious to say, that casting a JSON-document to an XML-document is done via `fn:json-to-xml(fn:serialize(., map{"method":"json"}))`.
1. Casting an XML-document to a text document seems to be definable via `fn:serialize(.)`.
1. At least for some XML-documents XPath 3.1 defines a conversion to JSON in `fn:xml-to-json()`. We could use this to define casting from (some) XML-documents to a JSON documents via: `fn:parse-json(fn:xml-to-json()).`

ndw:
Someone proposed that we were going to use p:cast-content-type to convert parameter sets into JSON.

xml-project:
Fixed with pr #285

PR #284, opened 2019-12-14
First sketch of p:cast-content-type

xml-project:
This is a first sketch for the revised "p:cast-content-type" (see #16). I tried to include @ndw's proposal to parse and serialize e.g. xml->text and vice versa. Also I try to come up for document property "serialization" which was not covered yet.
I do not expect an approve yet, but propose to discuss this on next editors call.


xml-project:
Try to fix error and come back.

PR #307, opened 2019-12-21
Fix simple typos in cast-content-type

ndw:
I fixed a few build warnings. The gloss terms used an emdash instead of a hypen in "implementation-defined"; this prevented the cross references from being found.

It turns out that all sections need to have `xml:id`s. The error if you put an `impl` in a section that doesn't have one was very odd!


ndw:
I assert that this is purely editorial.

PR #311, opened 2020-01-01
Clean up the specification of p:markdown-to-html

### LINKED ISSUE #39: p:markdown
I propose that we have a standard (optional) step for transforming markdown:

```xml
<p:declare-step type="p:markdown">
  <p:input port="source" content-types="text/*"/>
  <p:output port="result" content-types="application/xml+html"/>
  <p:option name="parameters" as="map(xs:QName,item()*)?"/>
</p:declare-step>
```

Parameters can be used to specify the processor and flavor of markdown.

ndw:
Fix #39 

I think this counts as a fix. We've moved the other text steps into the core step library, so this one can be for the markdown step. We might find more text-like steps to add later


ndw:
No, no, nothing like that. At the June workshop, we still had the text steps in this spec (I think) and the comment on `p:markdown` says "create a library for optional steps; or create a step library for this step."

All I meant in my comment above was that I think *this* document qualifies as a step library for this step.

xml-project:
@ndw 👍

PR #321, opened 2020-01-02
Attempt to fix-271

### LINKED ISSUE #271: XD0064 (invalid URI) must by raised in p:xslt
Has to be raised, iff the base-uri of a document appearing on the secondary port is not valid.

xml-project:
Fix #271
Created a new error code for this case as XD0064 is for options


xml-project:
> XSLT processor and the default or explicit base output URI will take care of it.

Nope: The XSLT processor is only required to check that result-document/href is an instance of XS:anyURI, not that it is a valid URI. This is why we need this additional check in the XProc processor.

gimsieke:
I was not talking about validity of the URI, only about making the URI absolute. I just wanted to make sure that users can continue writing `<xsl:result-document href="intermediate-results.xml">` without someone poking at them because they didn’t bother to use absolute URIs. It turned out that users may well continue to use relative hrefs.

xml-project:
@gimsieke Sorry, I missed your point.

PR #339, opened 2020-01-18
Fixed one of the To-be-done sections

xml-project:


steps/src/main/xml/steps/http-request.xml
```diff
@@ -230,10 +230,10 @@ For those methods the documents on the <port>source</port> are ignored when cons
           <varlistentry>
             <term>transfer-encoding</term>
             <listitem>
-              <note xml:id="http-uri-transfer-encoding" role="editorial">
-                <title>Editorial Note</title>
-                <para>TBD: How does this entry differ from the "transfer-encoding" header?</para>
-              </note>
+              <para>The value associated with the “<literal>transfer-encoding</literal>” key must be an instance
+                of <code>xs:string</code>. If the key-value pair is present, the response will be treated as if the response
+                contained a “<literal>Transfer-Encoding</literal>” header given by its value. The original Transfer-Encoding
```
gimsieke: Suggestion: contained a “<literal>Transfer-Encoding</literal>” header with the same value

Also for `Content-Encoding` in the preceding item.

steps/src/main/xml/steps/http-request.xml
```diff
@@ -230,10 +230,10 @@ For those methods the documents on the <port>source</port> are ignored when cons
...
+                will however be reflected unchanged as entry in the document on the <port>report</port> port.</para>
```
gimsieke: The original Transfer-Encoding will however be reflected unchanged as entry in the document on the report port. 
→
The original Transfer-Encoding *header value* will however be reflected unchanged as *an* entry in the document on the report port. 

Btw, we didn’t mention the `report` port in the spec. Instead, there is `response` port in “The response's status code and the response's header fields will be made available on the response port as a map.”. It should probably be changed into “`report` port”.

xml-project: The port "report" is named in the step's signature. But the line you found is surely a typo.

xml-project:
Since the changes requested by @gimsieke also apply to at least another sections (where parameter 'content-encoding') I propose you discuss this on Thursday's call and leave me a note about your resolution.

steps/src/main/xml/steps/http-request.xml
xml-project: Typo in port name addressed in pr #341

xml-project: ... and solved.

xml-project:
Overtaken by pr #346

PR #342, opened 2020-01-24
document properties of steps without primary inputs

### LINKED ISSUE #325: Document property preservation in the absence of an input port
The following steps state that “no document properties are preserved” despite the fact that there is no input port:

- [`p:www-form-urldecode`](https://spec.xproc.org/master/head/steps/#c.www-form-urldecode)
- [`p:www-form-urlencode`](https://spec.xproc.org/master/head/steps/#c.www-form-urlencode)
- [`p:directory-list`](https://spec.xproc.org/master/head/file/#c.directory-list) (also: add remark that `base-uri` is the URI of the directory)
- [`p:file-copy`](https://spec.xproc.org/master/head/file/#c.file-copy)
- [`p:file-delete`](https://spec.xproc.org/master/head/file/#c.file-delete)
- [`p:file-info`](https://spec.xproc.org/master/head/file/#c.file-info)
- [`p:file-mkdir`](https://spec.xproc.org/master/head/file/#c.file-mkdir)
- [`p:file-move`](https://spec.xproc.org/master/head/file/#c.file-move)
- [`p:file-file-create-tempfile`](https://spec.xproc.org/master/head/file/#c.file-create-tempfile)
- [`p:file-touch`](https://spec.xproc.org/master/head/file/#c.file-touch) (it is not specified what is contained in the result document if the step succeeds; also “one of the following errors” precedes a list of exactly one possible error)
- [`p:os-info`](https://spec.xproc.org/master/head/os/#c.os-info)

I’d say that the concept of property preservation only applies to steps with a (primary) input port.

Apart from the parenthesized remarks in the list above that might require special wording, a proposal for a general change of the document property sentence for these steps is something like this:

> The resulting document only has the `content-type` property. It does not have a `base-uri` property.

or 

> The resulting document has no properties apart from `content-type`. In particular, it has no `base-uri`.

gimsieke:
fix #325

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -279,7 +279,9 @@ steps is assumed; for background details, see
     
     <simplesect>
       <title>Document properties</title>
-      <para feature="directory-list-preserves-none">No document properties are preserved.</para>
+      <para feature="directory-list-preserves-none">Besides the <property>content-type</property> property, 
+        the resulting document has a <property>base-uri</property>. Its value is identical to the top-level
+        element’s <tag class="attribute">xml:base</tag> attribute, that is, to the directory’s URI.</para>
     </simplesect>
   </section>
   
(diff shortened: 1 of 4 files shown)


gimsieke:
ping

PR #346, opened 2020-01-29
Attempt to complete p:http-request

ndw:
Right. Here's my attempt to complete the `p:http-request` step. I made an editorial pass, hopefully my edits are acceptable. I made a few non-editorial change as well:

1. I removed the reference to certificates. I don't know how to write that section and I don't think we have to. We say that implementations may support additional auth-methods. Certificates can be one of them and we can come back to standardize later if we need to.
2. I renamed "multipart" to "accept-multipart"
3. I added "multipart-boundary"
4. I added "multipart-content-type"
5. I renamed the "content-encoding" parameter to "override-content-encoding" as I think it's parallel with "override-content-type".
6. I changed the "timeout" from decimal to integer because that seemed more consistent.
7. I attempt to write the sections about how multipart requests and responses are handled.

There a at least a half dozen or so places where we either explicitly or implicitly say "must" and haven't given an error code for what to do if that's not possible. I didn't try to fill all of those in as I don't think they're substantive.

Here's a [formatted draft](http://xpspectest.nwalsh.com/http-request-editorial/head/steps/#c.http-request)

Comments, etc. Most welcome.

### DIFF
--- a/src/main/xml/bibliography.xml
+++ b/src/main/xml/bibliography.xml
@@ -239,10 +239,16 @@ Andrew Coleman and C. M. Sperberg-McQueen, editors. W3C Recommendation. 21 March
 XML Format for mail and other messages</citetitle>.
 G. Klyne. Network Working Group, IETF, April 2002.</bibliomixed>
 
-<bibliomixed xml:id="rfc1321"><abbrev>MD5</abbrev>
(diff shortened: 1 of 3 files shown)


xml-project:
Great! Will have a look on it this afternoon.

ndw:
I agree about errors. And I agree that the prose needs another pass for them, thanks Achim!

PR #360, opened 2020-02-23
Attempt to fix p:file-touch

xml-project:


step-file/src/main/xml/specification.xml
```diff
@@ -405,8 +405,8 @@ steps is assumed; for background details, see
   <para>The <tag>p:file-info</tag> step returns information about the file, directory or other file system object named
     in the <option>href</option> option.</para>
 
-  <para><impl>Conformant processors <rfc2119>must</rfc2119> support file infos whose
-    scheme is <code>file</code>. It is
+  <para><impl>Conformant processors <rfc2119>must</rfc2119> support URIs whose
+    scheme is <code>file</code> for the <option>href</option> of <tag>p:file-info</tag>.
```
gimsieke: for the `href` _option_ of `p:file-info`
(I tend to omit the word “option”, too, because the tag is already called option)

gimsieke: And I think “It is”, the start of the following sentence, has been deleted.

step-file/src/main/xml/specification.xml
```diff
@@ -691,13 +691,23 @@ timestamp of the file specified in the <option>href</option>
...
+      implementation does not support p:file-touch for a specified scheme.</error>
+  </para>
+  
+  <para>If <option>href</option> is relative, it is made absolute against the 
+    base URI of the element on which it is specified 
+    (<tag>p:with-option</tag> or <tag>p:file-touch</tag> in the case of a 
+    syntactic shortcut value). <error code="D0064">It is a <glossterm>dynamic 
+      error</glossterm> if the base URI is not both absolute and valid according to <biblioref
+        linkend="rfc3986"/>.</error> <error code="C0137">It is a
+          <glossterm>dynamic error</glossterm> if p:file-touch can not perform due to access
```
gimsieke: can not → cannot

perform → be completed (my feeling is that unless it is in the passive voice, “to perform” always needs a target, either as a transitive verb or in the sense of performing in a play)

PR #370, opened 2020-03-18
Use document properties as headers only if they're in the correct namespace

### LINKED ISSUE #369: p:http-request is underspecified with respect to document properties as headers
The spec says:

> If a single document appears on the source port, then document properties on that document may be added as additional headers. All of the document properties will be used as headers except “base-uri”, “serialization” and any property...

I think we should say that only document properties in no-namespace are added. Otherwise, what do you do with a document property named `ex:foo`. You can't use "ex:foo" as a name. Even if `{` and `}` are allowed in header names (which I doubt), you can't use the `Q{}` syntax because the URI is almost always going to contain a colon.

We could say that you use the local name, but then you will always have errors if there are different properties with the same local name.

ndw:
Attempt to fix #369

I invented a new namespace; I wonder if using the step namespace would be better.

### DIFF
--- a/steps/src/main/xml/steps/http-request.xml
+++ b/steps/src/main/xml/steps/http-request.xml
@@ -492,7 +492,7 @@ are used to determine how serialization is performed.</para>
 <para>All of the document properties in the
 <code>http://www.w3.org/ns/xproc-http</code> namespace will be added
 as headers for the part, using the local-name of the property QName as
-the header name. In particular, this is now the
+the header name. In particular, this is how the
 “<literal>id</literal>”, “<literal>description</literal>”,
 “<literal>disposition</literal>” and other multipart headers can be
 provided.</para>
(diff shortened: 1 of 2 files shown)


steps/src/main/xml/steps/http-request.xml
```diff
@@ -487,11 +488,13 @@ assure that this cannot happen.</para>
...
 
-<para>All of the document properties <emphasis>except</emphasis>
-“<literal>base-uri</literal>” and “<literal>serialization</literal>”
-will be used as headers for the MIME part. In particular, this is now the
-“<literal>id</literal>”, “<literal>description</literal>”, “<literal>disposition</literal>”
-and other multipart headers can be provided.</para>
+<para>All of the document properties in the
+<code>http://www.w3.org/ns/xproc-http</code> namespace will be added
+as headers for the part, using the local-name of the property QName as
+the header name. In particular, this is now the
```
gimsieke: this is _how_ the …?

Issue #378, opened 2020-03-23
Better specification for filename regexp use


xatapult:
As @gimsieke noted in #364, the usage of regexps in filename matching is under-specified:

1. Make sure that is is clear everywhere that the regular expression is _not_ anchored. 
   * `p:directory-list`
   * `p:unarchive`
1. There is no error specified in the spec for when the regexp is invalid. Do we need to add a specific error code for that?

I stated in #364 that the matching can be described as using the XPath `matches#2()` function (as in `matches(path-to-match, regexp)`). Is that correct? If so, maybe we can describe it as such in the spec?

xatapult:
Consensus that this is correct and should be done

Issue #364, opened 2020-03-09
Reading XML files from an archive with weird filenames


xatapult:
I ran into a use-case for which we have no elegant solution in XProc 3.0 but is quite common. @gimsieke , I think this is rather important for transpect too!

MS Office files (like `.docx` and `.xlsx`) are zip files with all kinds of XML inside. Some have filenames like `.rels`. Since this is not a standard XML file extension, `p:unarchive` does not recognize it as XML and loads it as binary. But you need the XML to interpret the office format correctly. 

The base cause is of course that we load the contents of an archive in one go using `p:unarchive`. There is no mechanism to specify the content types of the individual files, you have to leave that to the processor which does its best but is sometimes wrong.

So how can we solve this? I see several solutions:

1. We do nothing. To convert the binary to XML, you'll have to write it to disk (to a temp file) and then read it in again. Works but is expensive and rather inelegant.
1. We add some feature to (probably) `p:cast-content-type` that forces it to view a binary file as text so you can convert it to XML after its loaded.
1. We add a mechanism to `p:unarchive` to help the processor guessing the file types. Maybe just a `content-type` attribute that forces all unarchived files to this type and then use `p:unarchive` to load the files one-by-one or in controlled sets using `include-filter` and `exclude-filter`.

Actually, my preference would be to implement both 2 and 3. Both are useful in their own way and probably other use-cases benefit also.

gimsieke:
My understanding was that processors can use any method that they deem appropriate in order to guess the correct media type. So I think they will at least support common XML archive members with unorthodox (or no) extensions.

In other functional languages one might solve this by submitting a function item to the step that accepts the file name and returns the content type.

Function items aren’t well supported in XProc 3.0, but maps are. So what about a `map(xs:string, xs:string)` with regular expressions as keys and content types as values? The semantics would be: If this map option is given and if a file name matches the regex supplied as a key, then the content type supplied as a value is returned, and the archive member is treated as if it were read with `p:load` with the respective `content-type` option. 

If a given file name matches multiple keys and these keys are associated with different content types, it is undefined which content type is selected.

Otherwise (if no key matches), the processor-dependent mapping rules apply that would apply in the absence of this map option.

xatapult:
I like Gerrit's solution. Sounds not too hard to implement. Opinions?

xatapult:
Is an extension to cast-content-type to treat binary as text if needed something to be considered as well?

[...]

PR #379, opened 2020-03-30
Specification for override content type

xatapult:
Issue #364 
Added error for invalid regular expressions. Will apply this error to other steps where applicable as of #378

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -279,7 +279,7 @@ draft:</para>
      <para>A <option>static-parameters</option> was added to <tag>p:xslt</tag>. (2020-03-23)</para>
    </listitem>
    <listitem>
-     <para>The <option>override-content-types</option> option was added to <tag>p:archive-manifest</tag> and <tag>p:unarchive</tag>. (2020-03-23)</para>
+     <para>The <option>override-content-types</option> option was added to <tag>p:archive-manifest</tag> and <tag>p:unarchive</tag>. (2020-03-30)</para>
    </listitem>
  </orderedlist>
 </appendix>
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -278,6 +278,9 @@ draft:</para>
    <listitem>
      <para>A <option>static-parameters</option> was added to <tag>p:xslt</tag>. (2020-03-23)</para>
    </listitem>
+   <listitem>
+     <para>The <option>override-content-types</option> option was added to <tag>p:archive-manifest</tag> and <tag>p:unarchive</tag>. (2020-03-23)</para>
+   </listitem>
  </orderedlist>
 </appendix>
 
--- a/steps/src/main/xml/steps/archive-manifest.xml
+++ b/steps/src/main/xml/steps/archive-manifest.xml
@@ -27,7 +27,7 @@
         archive. Implementations <rfc2119>must</rfc2119> support the <biblioref linkend="zip"/>
         format, specified with the value <code>zip</code>. <impl>It is
             <glossterm>implementation-defined</glossterm> what other formats are supported.</impl>
-        </para>
+      </para>
     </listitem>
     <listitem>
       <para>If no <option>format</option> option is specified or if its value is the empty sequence,
@@ -39,8 +39,9 @@
     </listitem>
   </itemizedlist>
 
-  <para><error code="C0085">It is a <glossterm>dynamic error</glossterm> if the format of the archive 
-    does not match the specified format, cannot be understood, determined and/or processed.</error></para>
+  <para><error code="C0085">It is a <glossterm>dynamic error</glossterm> if the format of the
+      archive does not match the specified format, cannot be understood, determined and/or
+      processed.</error></para>
 
   <para>The <option>parameters</option> option can be used to supply parameters to control the
     archive manifest generation. <impl>The semantics of the keys and the allowed values for these
@@ -50,10 +51,10 @@
       whose value is not valid for that key.</error></para>
 
   <para>The <option>relative-to</option> option, when present, is used in creating the value of the
-    manifest's <code>c:entry/@href</code> attribute. If the option is relative, it is made absolute 
-    against the base URI of the element on which it is specified (<tag>p:with-option</tag> or the step 
-    in case of a syntactic shortcut value). <error code="D0064">It is a <glossterm>dynamic 
-      error</glossterm> if the base URI is not both absolute and valid according to <biblioref
+    manifest's <code>c:entry/@href</code> attribute. If the option is relative, it is made absolute
+    against the base URI of the element on which it is specified (<tag>p:with-option</tag> or the
+    step in case of a syntactic shortcut value). <error code="D0064">It is a <glossterm>dynamic
+        error</glossterm> if the base URI is not both absolute and valid according to <biblioref
         linkend="rfc3986"/>.</error></para>
 
   <para>The generated manifest has the format as described in <xref linkend="cv.archive-manifest"/>.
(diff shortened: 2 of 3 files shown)


ndw:
What is the "relative filename of the archive's entry"?
Don't you just mean the filename of the archive's entry? I assume that if the archive filename is `/path/to/my/thing.xml`, I can write a regex that matches on `/path/to/my/` for example. Or did we decide otherwise?

xatapult:
AFAIK we decided to work on the relative filename in the archive, so it will never start with a /. This is also how its done in `p:unarchive`, for the include and exclude filters:

> Names of entries in archives are always relative names. For instance, the name of a file called xyz.xml in a specs subdirectory in an archive is called in full specs/xyz.xml (and not /specs/xyz.xml).

Whatever it is, it is worth making this clear (I tried by providing an example). It's always a nuisance these kinds of details.

ndw:
I'm fine with saying that we create archives with "relative" names, but I think in matching on unarchive, we should say we match against the actual filename in the directory entry. We shouldn't be tampering with it before applying the regex.

xatapult:
I don't follow. We're not "tampering". We're just defining exactly what string the regexp will match against. And whether that is with or without a leading `/`, I don't care, it's _just a choice_. As long as its defined unambiguously it's OK as far as I'm concerned.

I don't see any reason to change the decision we made on doing it against filenames that do not start with `/`, again, _just a choice_. But if you do, let's discuss this the next call. If we decide to change this we should IMHO do it everywhere, also in the include and exclude filters of `p:unarchive`.

@gimsieke , @xml-project Any opinions?

gimsieke:
I didn’t completely understand what kind of file name tampering Norm was afraid of, and I don’t think that Erik suggested any tampering. He was pointing out that all filters match against relative paths, without a leading slash. To my surprise, we seem to have gotten this consistent across the archive/unarchive and directory-list steps.

ndw:
All I'm saying is that we don't need to say anything about "relative paths" in this step. This step should match against the path specified in the archive, exactly as it appears in the archive. If someone used some tool that sticks a leading "/" on the archive filename, that should be part of what's matched against the regex. We shouldn't be removing leading slashes if they occur before we do the comparison.

xatapult:
> All I'm saying is that we don't need to say anything about "relative paths" in this step

Yes, we should. If you take an archive as the root of a file system, you need to define exactly how an entry in the archive is referenced. 

> If someone used some tool that sticks a leading "/" on the archive filename

I really don't follow. A file in an archive has a path leading up to it. How can you stick a slash in front?

Aren't we confusing the full path (the path of the archive itself + the path of the entry *in* the archive) with what we're dealing with here? What I mean is *only* the part *in* the archive.

gimsieke:
Even if someone managed to prepend a slash to each entry, this slash will be disregarded by `p:unarchive` and `p:archive-manifest` as per the two list items in front of [this message](https://spec.xproc.org/master/head/steps/#err.inline.C0120.1). So if we say that matching will be performed against the string in `c:entry/@name` (if someone requested an archive manifest), it will have no leading slash.

ndw:
I'm not sure what `relative-to` has to do with it. I'm also surprised that we seem to be talking past each other. Here's a zip file I have on my disk:
```
$ unzip -v example.zip
Archive:  example.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
       4  Defl:N        6 -50% 04-01-2020 16:05 7e3265a8  /tmp/x/readme
--------          -------  ---                            -------
       4                6 -50%                            1 file
```
I would expect this `override-content-type` to match:
```
[ [ '^/.*$', 'text/plain' ]]
```
Coming back to `relative-to`, suppose it's `file:///home/ndw/`. Then, off the top of my head, I expect that to work like this:
```
import java.net.URI
relative_to = URI.create("file:///home/ndw/")
relative_to.resolve("/tmp/x/readme")
```
And that returns `file:/tmp/x/readme`.

Granted, leading slashes are uncommon, a bad idea, and almost always wrong, but how is my analysis incorrect?

gimsieke:
`relative-to` doesn’t matter here as much as: “and with the relative path of this document as it was in the archive as first parameter”. In the resulting manifest your `/tmp/x/readme` will always appear as `<c:entry name="tmp/x/readme"/>`. This is because once you have the full URI, you need to strip away the leading `relative-to` or the leading slash of the path portion as per [the two items that precede this other message](https://spec.xproc.org/master/head/steps/#err.inline.C0118). 

I just noticed that what I tacitly assumed here is that before stripping away `relative-to`, a slash will be appended if there is no trailing slash. This is the behavior that `p:urify()` requires for its 2nd (`$basedir`) argument. 

We need to clarify this: 
1. If you supply `relative-to="file:///home/ndw"`, the full URI will be `file:///home/ndw/tmp/x/readme`.
1. If you supply `relative-to="file:///home/ndw/"`, the full URI will also be `file:///home/ndw/tmp/x/readme`.

So `c:entry/@name` of the resulting archive manifest should be `tmp/x/readme` in both cases. Therefore we need to say that the `relative-to`, potentially with a slash appended, will be subtracted from the full URI.

Alternatively, we can say that `c:entry/@name` for 1. will be `/tmp/x/readme` and for 2. it will be `tmp/x/readme`. But I think this alternative is not consistent with stripping away the leading slash for `c:entry/@name` in the following case:

> If there is no relative-to option or if its value is not the initial substring of the base URI of the document, the name is the path portion of the URI (per [RFC 3986]). If the path portion begins with an initial slash, that slash is removed.

xatapult:
Maybe it's just the term "relative" that's confusing?  

We can change that in: "the path leading up to the file in the archive, without a leading /"

xatapult:
I'm going to merge this PR (because its blocking another PR I want to make) but leave the issue open. I suppose the debate is not finished yet.

ndw:
@gimsieke writes:

> If you supply `relative-to="file:///home/ndw"`, the full URI will be `file:///home/ndw/tmp/x/readme`.

No, it won't. That is not how the RFCs define relative URI resolution. We should not be introducing magic that changes the established rules!

Also, if the filename in the archive is "/tmp/x/readme", the resolved  value will be `file:///tmp/x/readme` because that's also *how URI resolution works*.

gimsieke:
I understood that URIs of extracted archive members should _always_ (also for members with absolute paths) be `archive_uri/relative_path`. At least I remembered that this was our decision “because that’s how Java etc. form URIs for archive members” (Erik said that). 

This means that URI resolution needs to happen 

1. after a leading slash is stripped from the entry name and
2. after a slash is appended to the archive URI (or to the `relative-to` string) if there isn’t already a trailing slash

These rules make sure that an extracted member `/tmp/x/readme` of an archive `file:/home/ndw/example.zip` will get the URI `file:///home/ndw/example.zip/tmp/x/readme` (instead of `file:///tmp/x/readme`), which is indiscernible from the URI that a member with the name `tmp/x/readme` will have after extraction. 

I’m not saying that I want it to work this way, but I think this was the (maybe not fully informed wrt edge cases) consensus.

If this rule holds, it also means that unarchiving followed by archiving might not be idempotent: A member that had the name `/tmp/x/readme` will have the name `tmp/x/readme` after re-creating the archive (because “[when ZIP archives are processed, every name in the manifest must be a relative path without a leading slash](https://spec.xproc.org/master/head/steps/#cv.archive-zips)”, a statement that might also solicit further discussion).

If we go with “an extracted archive member `/tmp/x/readme` will have the URI `file:///tmp/x/readme` instead of `file:///home/ndw/example.zip/tmp/x/readme`, at least in the absence of `relative-to`”, then, in order not to render `relative-to` useless for absolute paths, we still need to stipulate that when `relative-to` is given, URI resolution still needs to happen like this:

1. after a leading slash is stripped from the entry name and
2. slash is appended the `relative-to` string, unless already present

Otherwise, as I said, `relative-to` will be useless if entry names are absolute paths.

PR #381, opened 2020-04-02
Issue 378

xatapult:
#378: Clarified the regular expression matching and added error code for invalid regexps.

We might get the same discussion here as for #379, but let's resolve that under issue #364


xatapult:
The regexp changes are not controversial (I hope). It's that the text mentions paths in archives which we might have to change. If we decide to do that we have to change a _lot_ more that's already in the spec. 

But I'm ok to wait.

xatapult:
@xml-project Could you (or somebody else) be persuaded to approve this PR? Then I can change the prose about paths in archives in one go (#385) after that, which will be easier for me.

Issue #380, opened 2020-03-30
Override content type mechanism in p:directory list?


xatapult:
Any objections against adding the (consensus) override content type mechanism #364 we designed for `p:unarchive` and `p:archive-manifest` to `p:directory-list` as well?

xml-project:
Let's do this as soon as we fixed the other issue. Wonder it it makes also sense on p:file-info

xatapult:
Consensus, yes.

PR #387, opened 2020-04-09
Added override-content-types option to p:directory-list and p:file-info

xatapult:
Prose for #380

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -91,6 +91,7 @@ steps is assumed; for background details, see
       <p:option name="max-depth" as="xs:string?" select="'1'"/>
       <p:option name="include-filter" as="xs:string*"/>
       <p:option name="exclude-filter" as="xs:string*"/>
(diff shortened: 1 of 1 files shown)


step-file/src/main/xml/specification.xml
```diff
@@ -473,6 +482,12 @@ steps is assumed; for background details, see
         if applicable.</para>
     </listitem>
   </itemizedlist>
+  
+    <para>The <option>override-content-types</option> option can be used to partially override the
+      content-type determination mechanism for files. This works just like with the
+        <option>override-content-types</option> option of <tag>p:archive-manifest</tag> and
+        <tag>p:unarchive</tag>, except that the regular expression matching is done against the
+      absolute URI of the file.</para>
```
gimsieke, line 490: Absolute URI or whatever string was given in the `href` option? I don’t have a strong opinion about it.

xatapult: Neither have I, just a choice

PR #393, opened 2020-04-20
Clarified override-content-types option 

xatapult:
Just added a reference for the `override-content-types` options in the file steps to where it is explained in the core steps library.

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -192,9 +192,9 @@ steps is assumed; for background details, see
     
     <para>The <option>override-content-types</option> option can be used to partially override the
       content-type determination mechanism. This works just like with the
-        <option>override-content-types</option> option of <tag>p:archive-manifest</tag> and 
-      <tag>p:unarchive</tag>, except that
-      the regular expression matching is done against the paths as used for the matching of the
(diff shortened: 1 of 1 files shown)


step-file/src/main/xml/specification.xml
```diff
@@ -486,8 +486,8 @@ steps is assumed; for background details, see
     <para>The <option>override-content-types</option> option can be used to partially override the
       content-type determination mechanism for files. This works just like with the
         <option>override-content-types</option> option of <tag>p:archive-manifest</tag> and
-        <tag>p:unarchive</tag>, except that the regular expression matching is done against the
-      absolute URI of the file.</para>
+        <tag>p:unarchive</tag> (see <xspecref spec="steps" xref="override-content-type"/>), except
```
gimsieke, line 489: I can’t judge whether these targets are correct. Is there supposed to be an `xml:id="override-content-type"` in some of the specs? To which of the source files does the XSL resolve `spec="steps"`?

xatapult: But of course there is. This is the way to create inter-spec-refs according to @ndw and it works. If you look in the steps spec you'll find a section with this id in p:archive-manifest

xatapult: This is how it looks like at a local build:

image

Issue #331, opened 2020-01-06
Specification of p:run


xatapult:
The description of the [p:run](https://spec.xproc.org/master/head/run/#c.run) step still starts with the text:

> THIS IS UNREVIEWED PLACHOLDER TEXT

I'm not sure what to make of this: Is the step description unfinished? Was it forgotten to remove this line?

xml-project:
I am pretty sure that the description is not finished yet, e.g. how to pass parameters to the step?

xatapult:
Let me give the specification of `p:run` a try and let's try to discuss some things before creating the final prose. I've looked at the Calabash [`cx:eval`](https://xmlcalabash.com/docs/reference/cx-eval.html) step for some inspiration, but in XProc 3, some things are rather different...

```xml
<p:declare-step type="p:run">
  <p:input port="source" primary="true" sequence="true" content-types="any"/>
  <p:input port="step" sequence="false" content-types="xml"/>
  <p:output port="result" primary="true" sequence="true" content-types="any"/>
  <p:option name="options" as="map(xs:QName, item()*)?"/>
  <p:option name="static-options" as="map(xs:QName, item()*)?"/>
  <p:option name="parameters" as="map(xs:QName, item()*)?"/>
  <p:option name="step-type" as="xs:QName?"/>
</p:declare-step>
```

- The `source` port connects to the primary port of the step-to-run, whatever its name. If there is no primary port on the step-to-run whatever appears on the `source` port is thrown away.
- The `step` port receives the step-to-run or a step library (see `step-type` option)
- The `result` port emits whatever comes out of the primary port of the step-to-run, whatever its name. If there is no primary output port on the step-to-run, it stays empty.
- The purpose of the options `options` and `static-options` are self documenting I think.
- `parameters` is for any additional processor dependent flags and parameter settings.
- `step-type` can specify the type of the step to run. It is mainly there to allow running a specific step from a library. If it is specified for a non-library step it must be the same as the type of the step. If the input is a library and `step-type` is absent, the first public step in the library is run.

This might be a basic setup. What we now cannot do is use additional ports. See next comment.

xatapult:
How to handle additional input and output ports on the step-to-run? Some options:

1. We do some syntax magic and allow connecting to ports on a `p:run` that are not defined. Nogo as far as I'm concerned. Feature creep.
1. We disallow them, only primary ports are allowed
[...]

PR #394, opened 2020-04-25
proposal for p:run, #331

gimsieke:
@ndw I referenced `<e:rng-pattern name="Run"/>` in specification.xml. The pattern probably needs to be defined by that name in core30.rnc in the grammar repo. Maybe you can change that after you merged https://github.com/xproc/3.0-grammar/pull/17

### DIFF
--- a/step-run/src/main/examples/run.xml
+++ b/step-run/src/main/examples/run.xml
@@ -1,0 +1,26 @@
+<p:run name="runme" xslt-parameters="{map{{'foo':'bar'}}}">
+  <p:with-input port="pipeline">
+    <p:inline expand-text="true">
+      <p:declare-step name="transform-n-validate">
+        <p:input name="source" primary="true" sequence="true"/>
+        <p:input name="stylesheet"/>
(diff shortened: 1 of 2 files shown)


gimsieke:
@ndw `../../../build/examples/run.txt` has not been created. I was naïvely hoping that it would automatically be generated if I added [`step-run/src/main/examples/run.xml`](https://github.com/gimsieke/3.0-steps/blob/run/step-run/src/main/examples/run.xml)

gimsieke:
@ndw I commented out `<e:rng-pattern name="Run"/>` for the time being and I included the example as escaped XML instead of XInclude, but the Travis build still fails:
```
> Task :step-run:rng FAILED
/home/travis/build/[secure]/step-run/build/steps.rnc:7:1: error: syntax error
```
The RNC of the run step should be in the 3.0-grammar repo that hasn’t changed because PR https://github.com/xproc/3.0-grammar/pull/17 hasn’t been merged.  

I tried to enable at least _some_ build so that one can look at the HTML rendering but I give up now.

PR #402, opened 2020-05-29
Attempt to resolve #395 (initial context item value)

### LINKED ISSUE #395: Change the default initial-context-item for p:xslt
Can we change the definition of the `initial-context-item` for `p:xslt` from:

```xml
<p:option name="global-context-item" as="item()?"/>
```

to:

```xml
<p:option name="global-context-item" as="item()?" select="."/>
```

If you don't do this it means that for global variables there is no context item. So you can't by default do something entirely normal (at least in my stylesheets) as `<xsl:variable name="..." select="/*/@something/>`. You would have to explicitly set the `initial-context-item` option to `.`.

This is also how Saxon behaves by default (when used directly from oXygen or from the command line), but it turns out to be a handy default set by Saxon, not something the spec defines. I think it's important to do this because people expect this when they run an XSLT 3 stylesheet: it is what they're used to.

xatapult:
Changes for issue #395 

@achim: I formulated the node-identity as: 

> Implementations **should** provide node-identity between the global context item and the document on the source port.

I think this is better than "implementation defined": It is now a recommendation but if a processor for whatever reason can't do it, it's free to not implement it. But if you think another way of saying it is more appropriate, let me know.

### DIFF
--- a/steps/src/main/xml/steps/xslt.xml
+++ b/steps/src/main/xml/steps/xslt.xml
@@ -47,11 +47,23 @@
     PSVI annotations.</para>
   <para>The interpretation of the input and output ports as well as for the other options depends on
     the selected XSLT version.</para>
+  
   <section>
     <title>Invoking an XSLT 3.0 stylesheet</title>
     <para>The value of <option>global-context-item</option> is used as global context item for the
-      stylesheet invocation. If no value is supplied, the empty sequence is supplied to the
-      invocation.</para>
+      stylesheet invocation. If no value is supplied, the following applies:</para>
+    <itemizedlist>
+      <listitem>
+        <para>If there is a single document on the <port>source</port> port, this document will
+          become the value of the <option>global-context-item</option> option. Implementations
+            <rfc2119>should</rfc2119> provide node-identity between the global context item and the
+          document on the source port.</para>
+      </listitem>
+      <listitem>
+        <para>If there are none or multiple documents on the <port>source</port> port, its value
+          will be the empty sequence.</para>
+      </listitem>
+    </itemizedlist>
     <para>If no value is supplied for <option>template-name</option> option an “Apply-template
       invocation” is performed. The documents that appear on <port>source</port> are taken to be the
       initial match selection. The default collection is undefined. If a value is supplied for 


xatapult:
Thanks @gimsieke.  However I would like @xml-project 's explicit approval given our discussion about node-identity.

xml-project:
@xatapult @gimsieke  I think it is missleading in two ways:

1. "Implementation" here means "implementations of XProc" but the node identity you are after, is node identity processing an XSLT stylesheet.
2. "should" is defined as

> that there  may exist valid reasons in particular circumstances to ignore a   particular item, but the full implications must be understood and   carefully weighed before choosing a different course. (RFC 2119)

My problem is the emphazis on "particular" while we are talking about the general case whether it is compliant to use an xslt processor for p:xslt that does not ensure node identity.

I think it is to much as the XSLT spec itself does not say anything about node identity between the node supplied for initial context item and the nodes in the initial match selection.

To my reading "Node identity is implementation defined" is a warning: Do not expect node identity also it might apply. Your suggestion means something completely different my reading.

Suggested compromise: Leave out any mention of node identity.

xatapult:
I do think we should mention the node-identity, because many XSLT developers will expect it (as I did too). I changed the wording to:

> Whether there will be node-identity between the global context item and the source document is implementation-dependent.

With the danger of being accused of riding a hobby-horse, I will add an issue to the core specification about node-identity and, don't worry, immediately declare it closed and deferred to 4.0. I think it was an oversight not to, at least, discuss this (not only for XSLT but node-identity in general). So if/when we ever get to a Vnext... we will not forget it.

xml-project:
Not happy with that, but I could agree.
@ndw Could you have a look at this please.

xatapult:
@xml-project Not happy with what exactly? I thought this was exactly what we agreed on in the last confcall...???

xml-project:
@xatapult As I said before: Node identity in an XSLT stylesheet is not something from which I think should be mentioned in an XProc specification at all.
But since other insists, I would agree. But that does not mean I have to be happy with it.

xatapult:
If it's still controversial, let's discuss it again next confcall, no need to rush. -------- Oorspronkelijk bericht --------Van: Achim Berndzen <notifications@github.com>Datum: zo 31 mei 2020 15:09Aan: "xproc/3.0-steps" <3.0-steps@noreply.github.com>Cc: Erik Siegel <erik@xatapult.nl>, Mention <mention@noreply.github.com>Onderwerp: Re: [xproc/3.0-steps] Attempt to resolve #395 (initial context item value) (#402)
@xatapult As I said before: Node identity in an XSLT stylesheet is not something from which I think should be mentioned in an XProc specification at all.
But since other insists, I would agree. But that does not mean I have to be happy with it.

xml-project:
@xatapult On the other hand: I said all I had to say about this, but it was not convincing. So let's see, what @ndw has to say about this and then go on.

xml-project:
Before merging pr #402: Please add an entry to the change log, so we can keep track of this.

ndw:
Specifying the global context item occurs at the API boundary between XProc and the XSLT processor, doesn't it? Wouldn't this be sufficient?
> If there is a single document on the <port>source</port> port, that document will
be specified as the value of the <option>global-context-item</option>.

If XProc passes the same node as the global context item and the document on the source port, then it's done all it can about node identity.

PR #405, opened 2020-06-09
Attempt to resolve spec issue 968

ndw:
This PR, along with corresponding changes in the specification repo is an attempt to close specification issue 968


ndw:
If you pass a step with PSVI annotations into p:identity, what you get out also has PSVI annotations.

xatapult:
Ehh, right. But what about steps that nowadays act like a `p:identity`, such as `p:store`? 

And what about core stuff like `p:if` with a false test?

What am I missing?

ndw:
You're not missing anything. I, on the other hand, wasn't paying attention. :(

PR #407, opened 2020-06-10
First draft to fix 403

xml-project:
This is a first draft of my ideas on how to make p:os-exec more XProc 3.0 like.
Will add the entry to the change log as soon as this gets approved. Not sure whether it would interfere with other PRs if I do it now.

### DIFF
--- a/step-os/src/main/xml/specification.xml
+++ b/step-os/src/main/xml/specification.xml
@@ -263,7 +263,7 @@ can be captured with a <tag>p:try</tag>.</para>
 <section xmlns="http://docbook.org/ns/docbook" xml:id="errors">
 <title>Step Errors</title>
 
-<para>This step can raise
+<para>These steps can raise
 <glossterm baseform="dynamic-error">dynamic errors</glossterm>.
 </para>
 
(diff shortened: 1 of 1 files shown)


xml-project:
Because I made many changes, a rendering might help: Please find it [here](https://xml-project.com/files/temp/os/)

xatapult:
It looks very good to me. For the use-case I have it would be more than sufficient. 

One small enhancement maybe: Define a default value of `<p:empty>`  for the `source` port?

cmarchand:
This sounds good to me, at least for the use-cases I have. For example an scp command...
And it's much more understandable than the [previous one](https://spec.xproc.org/master/head/os/#c.os-exec) !

xml-project:
@ndw @gimsieke Any objection to merging this and then sort out the minor questions about option naming?

xml-project:
@gimsieke Thanks for finding this. I just changed the section on p:os-exec, so I did not look at the rest.
Are we sure that we will not add additional steps to this library. If so and those steps raise errors, we would have to change it back again.
Suggestion: Leave it as section 4, but change singular to plural. Does that make sense?

gimsieke:
Yes, of course.

PR #414, opened 2020-06-20
Attempt to fix issue 409

### LINKED ISSUE #409: Encoding in header "Content-Type" of p:http-request
Currently we say for a single document to be send via HTTP (if there is no explicit deader 'Content-Type'):

> If a single document appears on the source port, then document properties on that document may be added as additional headers. 

However then the 'Content-Type' header constructed will not reflect any settings for 'encoding' in the 'serialization'.

We agreed to change this (Editors call 2020-06-10) to change that and to take the encoding into account when creating the default header.

Some questions:

1. Do we want this for XML document toos, so a standard XML document is send with header 'ContentType : application/xml; charset=UTF-8'?
2. I think for "other" (binary documents) we will not add 'charset' (whatever the serialization parameters say), because they are not serialized. Right?
3. For serialization method 'adaptive' encoding is not relevant, so I guess in this case 'encoding' is not used to construct the header 'Content-Type', right?

xml-project:
Attempt to fix #409 
Does this work?

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -314,6 +314,10 @@ draft:</para>
    <listitem>
      <para>Option <option>populate-default-collection</option> is added to the signature of <tag>p:xslt</tag>. (2020-06-20)</para>
    </listitem>
+   <listitem>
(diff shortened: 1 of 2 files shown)


xml-project:
@gimsieke @ndw @xatapult Any comment on this?

PR #432, opened 2020-07-06
attempt to fix 353

### LINKED ISSUE #353: File steps: Performance or state?
At least for two steps (make directory and delete file) the specs are open to different interpretations: Suppose I want to create a folder but I can‘t because the folder is already there. (Error or not). Same with delete files: I might be unable to delete a file/folder because it does not exist.
I propose to have a „state-interpretation“ saying: The author wants this folder to exist after the step finished, so it is not an error the call make-directory with an already existing Uri. In the same line: It is not an error to call p:delete-file with a uri of a non existing file.
Do you agree?

xml-project:
Close #353 by stating clearly that p:file-delete and p:mk-dir are idempotent

### DIFF
--- a/step-file/src/main/xml/specification.xml
+++ b/step-file/src/main/xml/specification.xml
@@ -594,8 +594,7 @@ steps is assumed; for background details, see
           in the environment in which the pipeline is run.</error></para>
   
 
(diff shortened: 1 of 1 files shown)


xml-project:
The PR currently does not pass the checks because Travis is unable to instantiate a JVM. I will try it again tomorrow morning.

PR #461, opened 2020-08-23
p:run-input, p:run-option etc.

gimsieke:


xml-project (approved):
Looks good to me. I think we should merge this, so we have a common base for discussions.
I will try to implement this and see, whether further question arise.
However, even after reading it more times, I fail to understand to meaning of the paragraph inline 120-121:

>The default readable port of the <tag>p:run</tag> step becomes the default readable port 
of the pipeline.

To my reading the DRP of p:run is the only input port declared which contains the pipeline. What did I miss?

gimsieke:
> To my reading the DRP of p:run is the only input port declared which contains the pipeline. What did I miss?

DRP ≠ primary input port 

https://spec.xproc.org/master/head/xproc/#dt-default-readable-port

> 3. If there is a preceding sibling step element:
>     If that preceding sibling has a primary output port, then that output port becomes the default readable port.
> 4. If there is not a preceding sibling step element:
>    If the container has a primary input port, the default readable port is that primary input port.

xml-project:
Thanks, @gimsieke 

1. It confused me that the quoted paragraph is between the `p:run-input` and the `p:run-option`. So I took it to be about inputs. Proposal: Move it to the section on p:output.
2. Do we need this sentence at all, because after all p:run has normal p:output declarations=

gimsieke:
Maybe there is still a misconception about DRP? It is not an output port of `p:run`, but of the preceding step.

It’s important to state that the primary inputs (and the options) of the dynamically executed pipeline receive the primary output of the step that precedes `p:run` (or the primary input of `p:run`’s container). 

Since the DRP affects inputs and options of the dynamically executed pipeline, it is placed appropriately where it is IMHO.

xml-project:
> It’s important to state that the primary inputs (and the options) of the dynamically executed pipeline receive the primary output of the step that precedes p:run (or the primary input of p:run’s container).

Ok, now I see what you are after. But I am not sure that this kind of magic is possible. And there should surely be more prose to explain this and its interference with p:run-input and p:run-option.

PR #463, opened 2020-08-24
Fixed example

xml-project:


gimsieke (approved):
Thanks for the corrections. 

`select="()"` isn’t strictly necessary since if there is no value given, option values default to the empty sequence

xml-project:
> Am 24.08.2020 um 15:35 schrieb Gerrit Imsieke <notifications@github.com>:
> 
> Thanks for the corrections.
> 
> select="()" isn’t strictly necessary since if there is no value given, option values default to the empty sequence
> 
Yap, but were are talking about a variable, don’t we? And @select is required on p:variable.
--------------------------------------
Achim Berndzen
achim.berndzen@xml-project.com

<xml-project /> Achim Berndzen
Kleine Breite 26a
38302 Wolfenbüttel, Germany

https://www.xml-project.com

gimsieke:
Ah ok, you’re right of course

Issue #462, opened 2020-08-24
please clarify relative-to option for p:archive


mkraetke:
In the spec it is stated that "The relative-to option is used in creating a manifest when no manifest is provided on the manifest port." For the reader remains unclear what an XProc implementation should do when both a manifest and an URI via `relative-to` is provided. IMHO, a dynamic error should be raised. Additionally, there is the text "If the option is relative". I think this must be "If the URI is relative".

xatapult:
@mkraetke Hi Martin, thanks for the comment. We will take it up in our XProc editors meeting this week.

xatapult:
Manifest has precedence

gimsieke:
Editors’ call today: Manifest has precedence

PR #469, opened 2020-08-28
Some small text changes for issue 462

xatapult:
Issue #462: Text changes/clarifications for the handling of the `relative-to` option

### DIFF
--- a/steps/src/main/xml/steps/archive.xml
+++ b/steps/src/main/xml/steps/archive.xml
@@ -123,10 +123,10 @@
       <term><option>relative-to</option></term>
       <listitem>
         <para>The <option>relative-to</option> option is used in creating a manifest when no
-          manifest is provided on the <port>manifest</port> port (if a manifest is present, the
-          option is ignored). If the option&#x2019;s value is a relative URI, it is made absolute against the base URI of
-          the element on which it is specified (<tag>p:with-option</tag> or the step in case of a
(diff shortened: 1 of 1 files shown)


xml-project (commented):
Sorry for going into subtle details. Those come up when you try to transform prose into code: "I would suggest to say, that the option's value is not used."
If we say that the option is ignored, a reader might interfer that he can give any value for `relative-to`. But it is defined as `xs:anyURI` and that has to be checked, whether the value is relevant or not. I am right on that?

xatapult:
Better. I'll change it! -------- Oorspronkelijk bericht --------Van: Achim Berndzen <notifications@github.com>Datum: vr 28 aug. 2020 12:24Aan: "xproc/3.0-steps" <3.0-steps@noreply.github.com>Cc: Erik Siegel <erik@xatapult.nl>, Author <author@noreply.github.com>Onderwerp: Re: [xproc/3.0-steps] Some small text changes for issue 462 (#469)
@xml-project commented on this pull request.

Sorry for going into subtle details. Those come up when you try to transform prose into code: "I would suggest to say, that the option's value is not used."
If we say that the option is ignored, a reader might interfer that he can give any value for relative-to. But it is defined as xs:anyURI and that has to be checked, whether the value is relevant or not. I am right on that?

xml-project (approved):

Issue #470, opened 2020-10-19
Output of empty p:unwrap


xatapult:
Recently I had the situation where a `p:unwrap` resulted in nothing: `<p:unwrap match="/*">` on a document with just an empty root element like `<some-root-element/>`.

To my surprise (before reading the specs) in Morgana it resulted in an empty document node or (I'm not sure) a document with an empty (?) text node. This is correct I suppose although the spec does not make this explicit. 

Can we make this situation more explicit and into something that makes more sense? I suggest to change the signature of the output port to `sequence="true"` and output empty in a case like described above.

gimsieke:
Hmm, not sure whether such a fringe case justifies changing the result to `sequence="true"`.

Did I understand it correctly: the only case that you think is not clearly specified if the result does not contain a text node at all? 

I think then the result should be a document node with no children. Empty text nodes are [not allowed](https://www.w3.org/TR/xpath-datamodel-31/#TextNodeOverview) in a parent node. But a document node may well [be empty](https://www.w3.org/TR/xpath-datamodel-31/#constraints-document). 

I’d read the spec that the content type of this document-node-only document is still `application/xml` or whatever the input document’s content type was.

xatapult:
Maybe you're right. Probably it's just a matter of explicitly mentioning and specifying this fringe case in the specs.

gimsieke:
And probably Morgana is right, too. I don’t have it installed, but can you inspect more closely the content type and `count(/node())` of the unwrapped document? It should be `application/xml` and `0`, respectively.

xatapult:
Unwrapping an empty element becomes a text document without contents.

ndw:
...per the editorial team meeting on 22 October 2020

xatapult:
I think I've opened a little can of worms with this. Even after the changes I already made, there are still some things left I think that should be specified or at least mentioned:

Gerrit mentions [here](https://github.com/xproc/3.0-steps/pull/471#issuecomment-715338215) that the `serialization` attribute will be removed when the content type changes. There is however no mention of that in the step's description. And I'm wondering: Is that necessary? Why not simply retain it, any no longer relevant serialization options ar simply ignored, right? If we *do* want to remove it, it should be mentioned in the step's description.

I was also wondering what would be the outcome of `<p:unwrap match="/*">` of:

```xml
<?Some processing instruction(s)?>
<!-- and/or some comment(s) -->
<an-empty-root-element/>
```

[...]

PR #471, opened 2020-10-23
Issue #470: Clarify p:unwrap for empty elements

xatapult:


ndw:
There's no such thing as an "empty text node", perhaps:

> &lt;para>This step produces a single document; if the document element is
unwrapped, the result might not be well-formed XML. If the document
element is empty or contains a single text node child, the result will
have the content type “&lt;literal>text/plain&lt;/literal>”. If the document
element is empty, the resulting document will be empty.&lt;/para>

xatapult:
Now I'm getting confused. I thought we decided that unwrapping an empty element would produce:

* A document node with
* A single text node child containing the empty string

Is this not correct?

gimsieke:
No, as Norm said above and I said in [#470](https://github.com/xproc/3.0-steps/issues/470#issuecomment-711756096), there is no such thing as an empty text node. As Norm suggested, when the unwrapped element doesn’t contain a node, the result will be an empty text document, that is, a document node with `content-type=text/plain` and no child nodes.

xatapult:
All right, I'll try to reformulate

xatapult:
Changed the last paragraph into a list and reformulated it a bit. Could you please check again?

gimsieke:
Shouldn’t we change the sentence “All document properties are preserved” to “With the exception of the special cases where `content-type` is changed to `text/plain` and a possible `serialization` property is removed, all document properties are preserved.”? Otherwise Ok!

xatapult:
@ndw Are you ok with Gerrit's suggestion?

ndw:
I'm ok with Gerrit's suggestion.

xatapult:
@ndw and/or @gimsieke I rephrased some sentences, to make it as clear as possible, so that even I can understand what's happening 😉 

Would you mind checking again?

gimsieke (approved):

PR #488, opened 2021-10-19
Add an error code for unsupported content types

ndw:


step-paged-media/src/main/xml/specification.xml
```diff
@@ -120,6 +120,8 @@ parameters on the <option>content-type</option> option is
 specified, the output type is
 <glossterm>implementation-defined</glossterm>.</impl> The default
 <rfc2119>should</rfc2119> be PDF.
```
gimsieke, line 122: Maybe: "should be PDF (`application/pdf`)."
Also in `p:xsl-formatter`.

ndw: No, I think what the spec is trying to say is that the default output format, in the absence of any other information, should be a PDF file. The `content-type` option doesn't have a default.

gimsieke: Ah ok

PR #497, opened 2022-04-27
Attempt to fix #490

### LINKED ISSUE #490: Use of `as="xs:token"` is an error
We have a number of steps that declare an option with enumerated values like so:

```xml
<p:option name="name" as="xs:token" values="('a','b','c')" select="'a'"/>
```

I believe this is an error. The default value `'a'` is a `xs:string`, `xs:string` is not a subtype of `xs:token`, and so the default value is statically an invalid type. That's bad. We could fix this with `value="xs:token('a')"`, but then you'd have to use `xs:token()` whenever you wanted to specify the value, so users would hate us. 😄 

Given that we're going to check the supplied value agains the list of values, there's no benefit in type checking against `xs:token`, or even `xs:string`, we should just remove the `as` type.

xml-project:
Changed as proposed.

### DIFF
--- a/steps/src/main/xml/steps/insert.xml
+++ b/steps/src/main/xml/steps/insert.xml
@@ -15,7 +15,7 @@ port's document relative to the matching elements in the
    <p:input port="insertion" sequence="true" content-types="xml html"/>
    <p:output port="result" content-types="xml html"/>
    <p:option name="match" as="xs:string" select="'/*'" e:type="XSLTSelectionPattern"/>
-   <p:option name="position" as="xs:token" values="('first-child','last-child','before','after')" select="'after'"/>
+   <p:option name="position" values="('first-child','last-child','before','after')" select="'after'"/>
 </p:declare-step>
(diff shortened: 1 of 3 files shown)


xml-project:
@ndw @gimsieke @xatapult 
There are no reviewers selected In the xproc/3.0-steps repo. Also: As a contributor I am not allowed to select a reviewer.
So this issue is just to inform you about new pull requests in case you are not informed automatically. You might want to fix reviewers too.
Cheers,
Achim

gimsieke (approved):

gimsieke:
I don’t know where to configure default reviewers for a repo. Can’t you manually select "Spec authors" as reviewer?

xml-project:
@gimsieke Nope: As a normal contributor you can't merge. select reviewers etc. The complete menu on the right side is greyed out. In the specs repo, reviewer "Spec authors" are preselected. May be @ndw knowns how to set this.

PR #499, opened 2022-06-13
Attempt to fix 486

### LINKED ISSUE #486: I'm unsure about err:XC0129
To save you looking it up:

> The http-version parameter indicates which version of HTTP must be used for the request. It is a dynamic error (err:XC0129) if the requested HTTP version is not supported.

I don't believe this is something the processor can know. Only the server knows what version(s) it supports. If there's a server out there that supports HTTP 2.0 or HTTP 3.0 or HTTP 99.9, who are we to refuse to send requests to it?

I believe the *server* is responsible for returning a 400 error if the requested version is not acceptable.

xml-project:
Attempt to fix #486 Error condition removed as proposed by @ndw


xatapult:
@ndw and @gimsieke I added you as reviewers (there was non registered). I feel out of my depth with regards to this issue.

ndw (approved):

PR #524, opened 2022-09-23
Attempt to fix 523

xml-project:


step-os/src/main/xml/specification.xml
```diff
@@ -92,9 +92,11 @@ steps is assumed; for background details, see
   </p:declare-step>
 
   <para>The step returns a <tag>c:result</tag> element with attributes
-  describing properties of the system. It <rfc2119>should</rfc2119>
-  include the following properties:</para>
-
+  describing properties of the system. It <rfc2119>must</rfc2119>
+  include the following properties. The attribute can be empty if they do not
+  apply to the run environment. For example <code>cwd</code> will be empty on systems
```
ndw, line 97: Idiomatically, I think "runtime environment" would be better than "run environment." And since I'm asking you to make a change :-), I suggest "the empty string" instead of just empty.

step-os/src/main/xml/specification.xml
```diff
@@ -153,7 +154,8 @@ steps is assumed; for background details, see
       environment variable. Implementations <rfc2119>should</rfc2119> include all available
       operating system environment variables.</para>
 
-  <para><impl>The exact set of properties returned is <glossterm>implementation-defined</glossterm>.</impl></para>
+  <para><impl>The exact set of properties returned is <glossterm>implementation-defined</glossterm>.</impl>
+  Additional attributes must be in a namespace.</para>
```
ndw, line 158: I think that should be marked as an `<rfc2119>` must.

ndw (approved):
I made a couple of small editorial proposals. I'm happy with the changes if you're happy to accept those comments.

xml-project:
@ndw Changes approved?

ndw:
Yes, sir.

ndw (approved):

PR #533, opened 2022-09-27
Attempt to fix 526

### LINKED ISSUE #526: Attribute `version` on dynamic pipeline in p:run
I think there is a problem with @version on the dynamic pipeline executed on p:run. If the attribute is present, all is fine. But if the attribute is absense, I think in the current writing the following rule from the main specs applies:

> If the p:declare-step has no ancestors in the XProc namespace, then it must have a version attribute. It is a static error err:XS0062 if a required version attribute is not present.

However once the `p:declare-step` arrives on the input port of p:run, it is almost impossible for the compiler to know whether it comes from an p:inline (no @version required) or is loaded from an external document via p:document or p:load (in some previous step).

Options:

1. The dynamic pipeline always get a missing version value from the pipeline which contains the p:run.
2. Attribute `version` is required for dynamically executed pipelines.

I could life with both of them and may be (1) is more convinient to authors.
Did I miss something?

xml-project:
Fix #526 by explicitly requiring a version attribute on pipeline root in p:run

### DIFF
--- a/step-run/src/main/xml/specification.xml
+++ b/step-run/src/main/xml/specification.xml
@@ -90,6 +90,8 @@ usual <tag>p:with-input</tag> and <tag>p:with-option</tag> connections.</para>
 
 <para>The <tag>p:run</tag> step
 expects a single pipeline document on its anonymous input port (“pipeline port”).
+Independed of the way the pipeline is provided, it needs to have attribute 
+  <literal>version</literal> on its root element. 
 <error code="C0200">It is a <glossterm>dynamic error</glossterm>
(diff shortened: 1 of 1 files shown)


ndw (changes requested):
I suggest instead:
```
Independent of how the pipeline is provided, it <rfc2119>must</rfc2119>
have a <tag class="attribute">version</tag> attribute on its root element.
```

ndw (changes requested):
"Independent" but otherwise fine. Thank you!  (With that typo fixed, feel free to merge.)

ndw (approved):
Sorry. I was looking at the wrong commit! *blush*

PR #535, opened 2022-10-01
Attempt to fix 534

### LINKED ISSUE #534: Asymmetry between input ports and options in p:run
If I do not miss anything important, I think there is an asymmetry while binding ports and options in p:run.
For ports the specs says:

> Input ports of the dynamically executed pipeline that are not declared with p:run-input on the p:run invocation will receive a p:empty connection. Input ports that are declared in p:run-input but not in the dynamically executed pipeline will be silently ignored.

And it is said for static options:

> If no such static option is declared by the pipeline or its imported libraries, the value passed in p:run-option will be silently ignored.

But there is no such rule for p:option (in the pipeline) and p:run-option (on p:run). 
I propose to have an equivalent rule for p:option/p:run-option:

> Options of the dynamically executed pipeline that are not declared with p:run-option on the p:run invocation will receive an empty sequence. Options that are declared on p:run-option but not in the dynamically executed pipeline will be silently ignored.

Did I miss something? Does this make sense?

xml-project:
Fix #534


ndw (approved):
"...will receive an empty sequence as *their* value." would be slightly more idiomatic English. Otherwise, I approve.

PR #541, opened 2023-03-25
Attempt to fix 540

### LINKED ISSUE #540: Signature of output ports of p:os-exec
Output ports "result" and "error" are not declared as sequence. But there are surely cases, were result or error is empty.
What to return in this case for the different values of "result-content-type". An empty std:out will not be a valid XML document, nor a valid JSON document, nor a valid Text document.
My expectation:  The port is empty, but then it has to be declared as sequence.
What did I miss?

xml-project:
Does this fix #540

### DIFF
--- a/step-os/src/main/xml/specification.xml
+++ b/step-os/src/main/xml/specification.xml
@@ -175,8 +175,8 @@ from standard error.</para>
 
 <p:declare-step type="p:os-exec">
   <p:input port="source" sequence="true" content-types="any"/>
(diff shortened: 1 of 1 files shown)


xatapult (approved):

xml-project:
@ndw @gimsieke would you mind to have a look at this pull request and either merge it or request changes. Thank you!

ndw:
Looks good to me. Apologies for the delay.

PR #558, opened 2024-04-24
Fix the sequence type of the default-version option

### LINKED ISSUE #556: Error in declaration for validate-with-json-schema
The `default-version` option is defined:

> ` <p:option name="default-version" as="xs:string"/>`

But the prose goes on to speak about what it means if the option isn't provided. If the option isn't provided, its value is `()` which raises a type error. The `default-version` should have the type `xs:string?`, I believe.

ndw:
(Assuming this will be an uncontroversial change.)

Fix #556 

Also added missing `impl` markup around the use of an implementation-defined glossterm.


gimsieke (approved):

ndw:
Close #556

PR #570, opened 2024-05-28
Improve the wording about matching document nodes

### LINKED ISSUE #566: Inconsistent definition of text/plain output from some steps
If you read the text of, for example, the `p:uuid` step, you'll find:

> If the document node is matched, the entire document is replaced by a text node with the UUID. What appears on port result is a text document with the text node wrapped in a document node.

Given the input:

```xml
<doc>Hello, world.</doc>
```

If the step specifies `match='/'`, the step is expected to produce a text document. Fair enough. But the *Document properties* section says:

> If the resulting document contains exactly one text node, the content-type property is changed to text/plain and the serialization property is removed, while all other document properties are preserved

As far as I can see, `match=doc` also produces a document containing exactly one text node.

There are tests in the test suite that require `match='doc'` to produce an `application/xml` document containing a single text node where `match='/'` must produce a `text/plain` document containing a single text node. 

It's quite practical to implement that, but I don't think the specification actually justifies that interpretation as currently drafted.

I don't think this is limited to the `p:uuid` step, I think it applies to `p:hash`, `p:replace`, `p:string-replace`, and perhaps others.

ndw:
Fix #566 

I think I've made the prose more parallel with the discussion about attributes and less likely to confuse men of small brain like me.

I also fixed a markup error in the step-validation specification and tweaked Gradle's daemon memory, but those should be irrelevant.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,3 +1,5 @@
+org.gradle.jvmargs=-Xmx4096M
+
 docbookXsltBaseUri=https://cdn.docbook.org
 docbookXsltVersion=2.5.0
 xprocTocUri=https://spec.xproc.org/master/head/etc/xproc/toc.xml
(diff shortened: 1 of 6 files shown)


ndw:
Ammendment: change the wording to "the result is a text document..." with a link to where text documents are defined.

Issue #563, opened 2024-05-04
Clarify p:set-attributes


ndw:
The `p:add-attributes` forbids adding attributes with the prefix `xmlns` or bound to `http://www.w3.org/2000/xmlns/`. Those same prohibitions apply `p:set-attributes`.

Also, the localname `xmlns` should also be forbidden.

PR #571, opened 2024-05-29
Clarify err:XC0059, add it to p:set-attributes

ndw:
I think reusing `err:XC0059` on `p:set-attributes` is the right thing to do...


ndw:
Close #563

xml-project (approved):

Issue #543, opened 2023-06-22
add <p:validate-with-dtd>


Markup-Fanatic:
**As a** XProc developer
**I want** native and performant DTD support for validation
**so that I** don't need to find a workaround by myself by implementig a <p:declare-step> with temporary storing, loading (and validating) and deleting my result

**Acceptance criteria**
- <p:validate-with-dtd> is provided

ndw:
Right. Okay. Achim and I chatted about this the other day.

DTD validation is *by definition* something that has to be performed on a sequence of characters. You can't "DTD validate" an XML document that's already been parsed. That's incoherent.

Here's a back-of-the-envelope design:

```xml
<p:declare-step type="p:validate-with-dtd">
     <p:input port="source" primary="true" content-types="xml html text"/>
     <p:output port="result" primary="true" content-types="xml"/>
     <p:output port="report" sequence="true" content-types="xml json"/>
     <p:option name="doctype-public" select="()" as="xs:string?"/>
     <p:option name="doctype-system" as="()" as="xs:string?"/>
     <p:option name="assert-valid" select="true()" as="xs:boolean"/>
     <p:option name="entities" as="map(xs:string,item())?"/>     
     <p:option name="report-format" select="'xvrl'" as="xs:string"/>
     <p:option name="serialization" as="map(xs:QName,item()*)?"/>
</p:declare-step>
```

If `entities` is non-empty, each entity must be either a string or a map containing a `doctype-system` key and an optional `doctype-public` key.

If the `source` is XML or HTML, it's serialized with a `<!DOCTYPE` declaration and declarations for the declared `entities` in the internal subset. If the `source` is text, it's the callers responsibility to format the text with a `<!DOCTYPE` declaration and other features.

The serialized document is parsed with an XML parser. The implementation may seralize the document "in memory" or on disk at its discretion. If `assert-valid` is true, a validating parser must be used. If `assert-valid` is false, a non-validating parser must be used. 

It is an error to provide `doctype-public` without `doctype-system`. It is an error to specify that `assert-valid` is true without providing a `doctype-system` value.

I'm not sure what to do about the base URI if the input doesn't have one. I guess you just get errors if any of the system ID references are not absolute.

This is probably incomplete, but it's a start.

xml-project:
>  If assert-valid is true, a validating parser must be used. If assert-valid is false, a non-validating parser must be used.

[...]

PR #579, opened 2024-06-07
First cut at p:validate-with-dtd

ndw:
This my first attempt. Feedback eagerly solicited. Formatted versions should appear on the xproc.org/dashboard page a few minutes after I create this request.

### DIFF
--- a/step-validation/src/main/xml/specification.xml
+++ b/step-validation/src/main/xml/specification.xml
@@ -510,119 +510,51 @@ to allow it to be optional.</para>
 
 <p:declare-step type="p:validate-with-dtd">
   <p:input port="source" primary="true" content-types="xml html text"/>
-  <p:input port="doctype" content-types="text" sequence="true"/>
+  <p:input port="doctype" content-types="text" sequence="true">
+    <p:empty/>
+  </p:input>
   <p:output port="result" primary="true" content-types="xml"/>
   <p:output port="report" sequence="true" content-types="xml json"/>
-  <p:option name="document-element" as="xs:string?"/>
-  <p:option name="general-entities" as="map(xs:string,item())?"/>     
-  <p:option name="external-entities" as="map(xs:string,map(xs:string,xs:string))?"/>     
   <p:option name="report-format" select="'xvrl'" as="xs:string"/>
   <p:option name="serialization" as="map(xs:QName,item()*)?"/>
+  <p:option name="assert-valid" select="true()" as="xs:boolean"/>
 </p:declare-step>
 
 <para>DTD validation differs from the other XML validation technologies in that
 it is applied during parsing. It isn’t possible to validate an XML data model with
 a DTD. This step necessarily serializes the source document and then parses it
-back into a new data model. The options on this step tailor the serialization.
+back into a new data model.
 </para>
 
-<para>The <tag>p:validate-with-dtd</tag> step does not have an
-<option>assert-valid</option> option. If validation fails, a new data model will
-not have been constructed. Consequently the step always fails if validation
-fails.
-<error code="C0210">It is a <glossterm>dynamic error</glossterm>
-if validation fails.</error>
+<para>There are several possible approaches, with varying degrees of complexity.
+The general model is that the contents of the <port>doctype</port> port and
+the result of serializing the <port>source</port> are concatenated together.
 </para>
-
-<para>The general form of the serialized result that will be DTD validated
-looks something like this:</para>
-
-<programlisting language="xml"><![CDATA[<!DOCTYPE root PUBLIC "-//Public//Identifier" "system-identifier.dtd" [
-<!ENTITY general "This is some text defined by a general entity">
-<!ENTITY external SYSTEM "chapter1.xml">
-]>
-<root>
-<p>Test: &general;.</p>
-&external;
-</root>]]></programlisting>
-
-<para>The first four lines are the document type declaration. What follows
-the declaration is the serialization of the source document. The declaration
-can be provided explicitly or generated.
-</para>
-
-<section>
-<title>Providing an explicit declaration</title>
-
-<para>Provide the entire declaration as a single text document on the
-<port>doctype</port> port.</para>
 
 <itemizedlist>
 <listitem>
-<para>The <option>general-entities</option> and
-<option>external-entities</option> options are ignored.
-</para>
+<para>In the simple case, the <port>doctype</port> is empty and <port>source</port>
+document is simply serialized. In order to have any chance of being DTD-valid,
+the serialization properties must include at least a <code>doctype-system</code>
+property.</para>
 </listitem>
 <listitem>
-<para>The serialization
-properties <code>doctype-system</code>, <code>doctype-public</code>, and
-<code>omit-xml-declaration</code> are ignored.
-</para>
+<para>If an internal subset is required, it is provided on the <port>doctype</port> port.
+In this case, the <code>source</code> document must be serialized <emphasis>without</emphasis>
+a <code>doctype-system</code> property; both the internal and external declarations must appear
+on the <port>doctype</port> port.</para>
 </listitem>
 <listitem>
-<para>The caller must assure that the root element listed in the
-<code>!DOCTYPE</code> declaration matches the root element of the source
-document.</para>
+<para>Finally, if a text document is provided on the <port>source</port> port,
+it is simply concatenated with the <port>doctype</port> port.</para>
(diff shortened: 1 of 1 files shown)


ndw:
Close #543

step-validation/src/main/xml/specification.xml
```diff
@@ -498,6 +503,127 @@ No document properties on the <port>schemas</port> port are preserved.</para>
...
+
+<para>If <option>external-entities</option> are provided, they must be
+a map from entity names to a map that identifies the system identifier and
+may identify a public identifier. For example,</para>
+
+<programlisting><![CDATA[map { 'chapter1': map { "system-identfier": "chap1.xml" },
+      'biblio': map { "public-identifier": "-//Bibliograph//EN",
+                      "system-identifier": "bib.xml" }}]]></programlisting>
+
+<para>The <code>system-identifier</code> property must be provided. The
```
gimsieke, line 609: Unless the full declaration is given on the `doctype` port, I guess?

step-validation/src/main/xml/specification.xml
```diff
@@ -498,6 +503,127 @@ No document properties on the <port>schemas</port> port are preserved.</para>
...
+
+<para>DTD validation differs from the other XML validation technologies in that
+it is applied during parsing. It isn’t possible to validate an XML data model with
+a DTD. This step necessarily serializes the source document and then parses it
+back into a new data model. The options on this step tailor the serialization.
+</para>
+
+<para>The <tag>p:validate-with-dtd</tag> step does not have an
+<option>assert-valid</option> option. If validation fails, a new data model will
+not have been constructed. Consequently the step always fails if validation
```
gimsieke, line 530: Is this really all that we can offer? I’d envisaged that it works like `xmllint --dtdvalid` (do a posteriori validation against a given DTD).

root.xml:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE root PUBLIC "-//Public//Identifier" "system-identifier.dtd">
<root>
  <a></a>
  <c></c>
</root>
```

system-identifier.dtd:

```dtd
<?xml version="1.0" encoding="UTF-8"?>
<!ELEMENT root (a, b)>
<!ELEMENT a (#PCDATA)>
<!ELEMENT b (#PCDATA)>
```

Invoke validation:

```
$ xmllint --noout --dtdvalid system-identifier.dtd root.xml
root.xml:3: element root: validity error : Element root content does not follow the DTD, expecting (a , b), got (a c )
root.xml:5: element c: validity error : No declaration for element c
Document root.xml does not validate against system-identifier.dtd
```

And then wrap the error message line into appropriate elements for the requested report format.

When I use Calabash and `p:load[@dtd-validate='true']`, I get something like this:

```
Error on line 5 column 6 of root.xml:
  SXXP0003   Error reported by XML parser: Element type "c" must be declared.: Element type
  "c" must be declared.
Error on line 6 column 8 of root.xml:
  SXXP0003   Error reported by XML parser: The content of element type "root" must match
  "(a,b)".: The content of element type "root" must match "(a,b)".
<c:errors xmlns:c="http://www.w3.org/ns/xproc-step"><c:error xmlns:err="http://www.w3.org/ns/xproc-error" code="err:XC0027" href="file:/mnt/c/Users/gerrit/XML/XProc/2024-06_validate-with-dtd/validate-with-dtd.xpl" line="3" column="59">The XML parser reported two validation errors</c:error></c:errors>
```

The lines before `c:error` are just written to STDOUT. I think they’d need to be collected instead and put into the report, each message wrapped into something like

```xml
<detection severity="error" code="SXXP0003">
  <location  line="5" column="6"/>
  <message xml:lang="en"> Element type "c" must be declared.</message>
</detection>
<detection severity="error" code="SXXP0003">
  <location  line="6" column="8"/>
  <message xml:lang="en">The content of element type "root" must match "(a,b)"</message>
</detection>
```

step-validation/src/main/xml/specification.xml
```diff
@@ -498,6 +503,127 @@ No document properties on the <port>schemas</port> port are preserved.</para>
...
+</section>
+<section>
+<title>Validation</title>
+<para>After the doctype declaration has been constructed, the source document
+is appended to it. If the source is XML or HTML, it will be serialized. Text
+documents are just appended to the declaration.</para>
+<para>The resulting text is parsed using a validating XML parser.</para>
+
+<para>Any warning messages produced by the parser will appear on the
+<port>report</port> port.</para>
```
gimsieke, line 623: See above, I’d also expect the actual validation errors to be listed in the report.

ndw:
Thank you, Gerrit! You're absolutely right. The step can return the original document unchanged if there was an error. That's much more sensible.

ndw:
I think we need a `document-element` option for the case where you send a text document as the source. For example:

```xml
<p:identity>
  <p:with-input><p>Paragraph of text.</p></p:with-input>
</p:identity>

<p:validate-with-dtd
  general-entities="map { 'text': 'Hello, world.',
                          'para': . }"
  document-element="doc">
  <p:with-input port="source">
    <p:inline content-type="text/plain"><![CDATA[<doc>
<p>Test</p>
<p>&text;</p>
&para;
</doc>]]></p:inline>
  </p:with-input>
  <p:with-input port="doctype"><p:empty/></p:with-input>
</p:validate-with-dtd>
```

ndw:
Having poked at the implementation a bit, I think what I've proposed is way over-the-top. How about:

```xml
<p:declare-step type="p:validate-with-dtd">
  <p:input port="source" primary="true" content-types="xml html text"/>
  <p:input port="doctype" content-types="text" sequence="true">
    <p:empty/>
  </p:input>
  <p:output port="result" primary="true" content-types="xml"/>
  <p:output port="report" sequence="true" content-types="xml json"/>
  <p:option name="report-format" select="'xvrl'" as="xs:string"/>
  <p:option name="serialization" as="map(xs:QName,item()*)?"/>
  <p:option name="assert-valid" select="true()" as="xs:boolean"/>
</p:declare-step>
```

1. The simple case, you pass a document with a `doctype-system` serialization property (on the document or the step). We serialize the document with the necessary doctype declaration and validate it.
2. You provide a `doctype`, we serialize the `source` document (without a doctype declaration or XML declaration), slap the `doctype` you provided in front of it and validate it.
3. If you want to do anything funky with entity replacements or some such, you construct the text of the document you want to parse, by whatever means you want, and we validate it.

xml-project:
Most probably missed something important, but I am confused what the report result port is for. If the validation succeeds, nothing “interesting” is in the documents on this port. If it doesn’t, the report document is not available because a dynamic error is raised.
What do I miss?

ndw:
Several comments back, @gimsieke persuaded me that we should put the `assert-valid` option back and just pass the original document through if `assert-valid` is `false()` and an error occurs.

xml-project:
@ndw thanks. Now I know what I missed. :-))

xml-project:
@ndw Two questions came up, while trying to implement the new suggestion:
```
<p:declare-step type="p:validate-with-dtd">
  <p:input port="source" primary="true" content-types="xml html text"/>
  <p:input port="doctype" content-types="text" sequence="true">
    <p:empty/>
  </p:input>
  <p:output port="result" primary="true" content-types="xml"/>
  <p:output port="report" sequence="true" content-types="xml json"/>
  <p:option name="report-format" select="'xvrl'" as="xs:string"/>
  <p:option name="serialization" as="map(xs:QName,item()*)?"/>
  <p:option name="assert-valid" select="true()" as="xs:boolean"/>
</p:declare-step>
´´´
Please excuse this questions, if they are stupid, but I am not a DTD-expert.
1. What is supposed to happen, if a Text document appears on port "source"?
2. Is an HTML document appears on port "source", is the result type "xml" correct?

**Comment by ndw** [comment]
A text document is allowed so that you could construct something like this:

```
<doc>
  &chap1;
  &chap2;
</doc>
```

where presumably the `chap1` and `chap2` entities are defined in the `doctype`. There's no way to get unexpanded entities into a parsed XDM, so you'd have to do it this way. I haven't thought very hard about how difficult it will be to make a text document that serializes correctly!

DTD validation sort-of implies XML, so I think making the result always be XML makes sense. If you think it makes more sense to give a document with a root element of (X)HTML an HTML content type, I can see how that might make sense too.

ndw:
Hi folks. I've pushed an update that simplifies the `p:validate-with-dtd` step along the lines that I described in a comment above.

gimsieke (approved):

PR #582, opened 2024-06-13
Prepare for 3.1 publication

ndw:
1. Update the publication metadata for 3.1
2. Replace the changelog with "changes since 3.0"
3. Removed the erratum sections added when we were considering doing a 3.0 with errata publication


xml-project (commented):
I think we are now ready for 3.1 publication, aren't we?
I think the request for the two dfdl-steps should got to XProc next because it will take some time to work this out.

ndw:
Yes. I think we're pretty close. I just pushed a fix for 1104 in the specification repo. I'll see about updating the 3.1 PRs then I'll merge them, I guess

PR #584, opened 2024-06-20
Attempt to fix issue 583

### LINKED ISSUE #583: Content-types of ports "source" and "result" on p:insert: Allow text documents?
I think there is an edge case, where the result of p:insert can be a text document. If the document on port "source" is an XML document which contains just a document node (no element nodes), and then I insert a text document, I think I should get a text document.
But the current specification says, port "result" on p:insert is either XML or HTML.

Proposal: Add "text" to the possible content types of port "result".

Does this make sense?

xml-project:
Attep to fix issue #583 
Close #583


ndw:
I'm happy for you to update this PR after we've resolved the issue, but if you'd rather close it, that's ok too.

PR #589, opened 2024-07-30
Fix #587 by adding a parameters option

### LINKED ISSUE #587: Add a parameters map to p:uuid?
It strikes me as slightly weird that the `p:uuid` step says:

>Implementations must support version 4 UUIDs. Support for other versions of UUID, and the mechanism by which the necessary inputs are made available for computing other versions, is [implementation-defined](https://spec.xproc.org/master/head/steps/#dt-implementation-defined).

Wouldn't it make sense to add a `parameters` map option so that there's a **place** to provide the necessary inputs? Implementors could use extension attributes or something, I suppose, but it seems like there's plenty of precedent for having a `parameters` map for this purpose...

ndw:
This gives the user somewhere to put parameters for other forms of UUID. They remain implementation defined.


gimsieke (approved):

ndw:
Close #587

PR #592, opened 2024-07-31
Restructuring the step summaries

ndw:
I was tinkering with a step last night and got a bee in my bonnet about the style of the step summaries. My first thought was that adding some syntax highlighting would improve things. And I guess maybe it did.

Then this morning, I got distracted by the idea that maybe the whole thing could be improved and made more readable.

So here it is. It's certainly different and...I think...it's an improvement.

One observation: the links from the names of ports and options in the summary to their descriptions only work if the descriptions are formatted as a variable list. The fact that we haven't done that in a lot of places is interesting. 

This relates to @xatapult 's thoughts about refactoring the spec entirely in some future version. But in this case there's no change in the spec, just a change in the presentation.

### DIFF
--- a/src/main/resources/css/xproc.css
+++ b/src/main/resources/css/xproc.css
@@ -184,3 +184,108 @@ sup.xrefspec a:visited {
 
 .assert {
 }
+
+/* ============================================================ */
+
+:root {
+    --tableaux-border-color: #aaaaaa;
+}
+
+div.declare-step {
+    padding-left: 1rem;
+    padding-right: 1rem;
+    padding-top: 0.5rem;
+    padding-bottom: 0.5rem;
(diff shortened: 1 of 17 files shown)


ndw:
It's an annoying "security feature" of the PR builds that stylesheet changes aren't reflected in them. See [this copy](https://nwalsh.com/scratch/xproc-style/steps/) to preview the style changes.

gimsieke (approved):
Yes, it is an improvement!

ndw:
I've added the implementation defined/dependent features to the summary and done a better brute-force job of linking from the names of ports and options to their descriptions.

gimsieke (approved):
nice!

PR #602, opened 2024-08-01
Fix #583 by allowing the result to be text

### LINKED ISSUE #583: Content-types of ports "source" and "result" on p:insert: Allow text documents?

ndw:
Editorial change: I replaced "The required steps" with "Step library" because we no longer have any /optional/ steps in this specification.

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -204,7 +204,7 @@ PSVI annotations.</para>
 </section>
 
 <section xml:id="req-steps">
-<title>The required steps</title>
+<title>Step library</title>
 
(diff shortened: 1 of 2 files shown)


ndw:
I'm assuming this is non-controversial because the issue has been open for a while and no one has objected. I realize I've made myself a merge conflict so I want to merge this and sort that out.

PR #603, opened 2024-08-01
Fix markup error

ndw:
Oops.


ndw:
Actually a stylesheet fix was needed. Which won't work in a PR. They're all editorial, I'm just going to merge it.

Issue #547, opened 2023-11-28
Missing error conditions for p:ixml


xml-project:
To my reading there are two missing error conditions in p:ixml

1. Port "grammar" is a sequence port which (according to the prose) can either be empty or provide one document. It is not said what happens if more than one document appears on this port. Proposal: It is a dynamic error if more than one document appears on port "grammar".
2. The document on port "grammar" may not be a valid ixml grammar. Proposal: It is a dynamic error if the document provided on port "grammar" is not a valid ixml grammar.

Does this make sense? Currently my implementation raises XD0030 (general step error) but I think we should be more precise here, especially in the second case.

PR #604, opened 2024-08-01
Improvements to p:ixml

### LINKED ISSUE #494: p:ixml should always wrap output into ixml element
I think that the principle of least surprise ([POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)) enforce the idea that the output should be consistent whatever the context we're in

At the moment, the specification as it is gives different ixml or no wrapper on different context
* example 1 : ixml wrapper
* example 2 : no wrapper
* example 3 : ixml wrapper
* example 4 : no wrapper
* example 5 : no wrapper
* example 6 : ixml wrapper

I would recommend that example 2, 4 and 5 to add a wrapper to avoid surprises

ndw:
For some reason, the build didn't include `p:ixml`. I fixed that.

I fixed #494 and #547.

I changed the content types of the `source` port given the new error changes.

### DIFF
--- a/step-ixml/build.gradle
+++ b/step-ixml/build.gradle
@@ -5,7 +5,7 @@ repositories {
 
 configurations {
   tools {
-    description = "ixml tools"
+    description = "Run tools"
     transitive = true
   }
 }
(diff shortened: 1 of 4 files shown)


step-ixml/build.gradle
```diff
@@ -5,7 +5,7 @@ repositories {
 
 configurations {
   tools {
-    description = "ixml tools"
+    description = "Run tools"
```
gimsieke, line 8: I assume this is just a human-processible info without any technical implication? (My first thought was that it was erroneously copied from `p:run` but now I think that it just says what it says.)

gimsieke (commented):
Is there a current rendering of the changed spec? [The one on xproc.org](https://spec.xproc.org/master/head/ixml/) dates from Nov 27, 2023.

ndw:
Oh, bah humbug. This is another instance of a build change not reflected in a PR. I'll merge it and we can blame me if I borked int.

PR #624, opened 2024-09-06
Attempt to clarify p:error

### LINKED ISSUE #623: Unspecified p:error behavior
The `p:error` step has a `source` port with `sequence="true"`. The spec leaves it unspecified what must happen when there are multiple documents on the `source` port. There are, I think, two options:

1. Use a separate `c:error` element for each incoming document
2. Add all incoming documents as children of a single `c:error` element.

To me the first makes more sense. It would allow you to raises multiple errors in one go. But it is not a very strong preference.

I would like to hear opinions and I think we should specify this.

ndw:
I took a closer look at `p:error` and thought it could be more clearly described. Hopefully this is an improvement...

Fix #623

### DIFF
--- a/steps/src/main/xml/steps/error.xml
+++ b/steps/src/main/xml/steps/error.xml
@@ -14,16 +14,18 @@ step.</para>
     <p:option name="code" required="true" as="xs:QName"/>
 </p:declare-step>
 
-<para>The <tag>p:error</tag> step always fails. It generates the error specified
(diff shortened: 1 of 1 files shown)


ndw:
In an offline conversation, @xml-project pointed out that making the results implementation-defined introduced the possibility of interoperability issues. I think multiple source documents are so unlikely, I was persuaded to just mandate an answer.

gimsieke (approved):

PR #628, opened 2024-09-11
Fixed 1.0 left-over in p:filter

xatapult:
I found a small 1.0 left-over in `p:filter` (mentioning `p:input` instead of `p:with-input`). This should fix it.


steps/src/main/xml/steps/filter.xml
```diff
@@ -14,7 +14,7 @@ based on a (possibly dynamically constructed) XPath select expression.</para>
   <p:option name="select" required="true" as="xs:string" e:type="XPathExpression"/>
 </p:declare-step>
 
-<para>This step behaves just like an <tag>p:input</tag> with
+<para>This step behaves just like an <tag>p:with-input</tag> with
```
ndw, line 17: I think you want `...just like a <tag>...` here. Not "an" because "with" doesn't begin with a vowel.

PR #645, opened 2024-12-24
Allow p:filter to return non-node results

### LINKED ISSUE #644: More problems with p:filter?
Back in [#79], @xml-project observed that the content type property on the `result` port for `p:filter` was wrong. We fixed it.

In the same issue @Conal-Tuohy raised a question about JSON results that we apparently missed. I think he's right. I think the output of the step should include JSON. Or maybe should be unconstrained.

If this filter (from the test `nw-filter-011`) is valid, which we agree it is:

```
         <p:filter select="/book/*[local-name(.) {$eq} '{$select}']"/>
```

then I can't think of any compelling reason why this isn't also valid:

```
         <p:filter select="count(/book/*[local-name(.) {$eq} '{$select}'])"/>
```

It's true that the step is described thus:

> The p:filter step selects portions of the source document based on a (possibly dynamically constructed) XPath select expression.

And maybe `count()` doesn't return "a portion" of the input. But I think what we *really* meant by `p:filter` was that it was a dynamic version of `select` on `p:with-input`.

ndw:
Fix #644

### DIFF
--- a/steps/src/main/xml/steps/filter.xml
+++ b/steps/src/main/xml/steps/filter.xml
@@ -5,24 +5,29 @@
          xmlns:xlink="http://www.w3.org/1999/xlink" xml:id="c.filter">
 <title>p:filter</title>
 
-<para>The <code>p:filter</code> step selects portions of the source document
-based on a (possibly dynamically constructed) XPath select expression.</para>
+<para>The <code>p:filter</code> step evaluates an XPath expression
+against the input document.</para>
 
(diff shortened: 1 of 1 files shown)


xml-project (approved):
The PR title is a bit confusing: The change does not only allow atomic values but also maps and arrays. But I think that is what we want.

ndw:
Yeah. That was careless of me. I fixed the title. Well, I think I improved it, anyway 😄

PR #649, opened 2025-01-05
p:sleep, clarify duration is in seconds, allow xs:dayTimeDurations

ndw:


xatapult:
I think it will be helpful to add an example to the already present table with a duration in milliseconds?

xatapult:
Looking at the milliseconds example you just added: I thought (coming from how it was) that the duration should be stated as an (non-negative) *integer* or an xs:dayTimeDuration. You now say that it can also be something with decimals. That makes absolute sense, but maybe the text should also specify the type of number that is acceptable ("can be cast as")  (xs:double? xs: decimal?). Probably xs:double?

ndw:
I tried to finesse that by just saying "number". In reality, I expect implementations to simply attempt to convert the string to a number (a float, a decimal, a double, whatever is convenient) and object if that fails. I don't see any value in requiring implementations to check what *kind* of number it is.

xatapult:
Hmm. This would mean that the kind of numbers accepted here are implementation-defined. What if I wrote 1E2 and one implementation accepted doubles and the other one didn't?

So I do think something like "castable as xs:double" should be added. xs:couble covers probably everything you're ever going to need as a number without limiting you  (or am I missing something?). And it doesn't harm, doesn't it?

Any other opinions?

xatapult (approved):

PR #663, opened 2025-02-01
Added the attributes option to p:pack, p:wrap, and p:wrap-sequence

### LINKED ISSUE #656: Adding attributes on element added with p;wrap
I think it should be possible to add attributes to elements added with p:wrap in a single step.  I guess that should be a pretty common requirement. Of course, you can always add the attributes in another step, but if the name of the added element is not unique in a file, you will need workarounds like adding the element with a temporary name first, then add the attribute, and finally rename the element again. That works, but it's a bit awkward.

ndw:
Fix #656

Incidentally, changed the order of p:wrap and p:wrap-sequence so that they appear in lexicographic order in the specification.

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -254,8 +254,8 @@ PSVI annotations.</para>
 <xi:include href="steps/uncompress.xml"/>
 <xi:include href="steps/unwrap.xml"/>
 <xi:include href="steps/uuid.xml"/>
-<xi:include href="steps/wrap-sequence.xml"/>
 <xi:include href="steps/wrap.xml"/>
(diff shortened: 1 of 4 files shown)


xml-project (approved):

xml-project (commented):
- What about adding namespace attributes? I think this should not work. So the section from "p:set-attributes" starting with "This step cannot be used to add namespace declarations. " needs to be added here. Does this make sense?
- Does addition an attribute "xml:base" change the base-uri? To be consistent I think it should. Do we need to say this here?

ndw:
Yes, on both counts, I think. I noticed the namespace attributes part when I was implementing it.

xml-project:
@ndw Same here!

PR #664, opened 2025-02-01
Added the p:encode step

### LINKED ISSUE #659: Create base64?
Is there an obvious way to create a base64 encoded version of a document?

It looks like you might be able to get there by going around the houses: getting the output serialized, somehow getting it's content type changed to something binary, then casting it to XML. 

But is there an easy way I'm just overlooking?

ndw:
Fix #659

In order to make the p:cast-content-type step decode any c:data document, even one that might contain serialized XML, I’ve reorganized the “Casting from an XML media type” section. The case of a c:data document is handled first.

This is a small backwards incompatibility because it means that you cannot use p:cast-content type to change the content type of a c:data element from application/xml to, for example, application/foo+xml. But that seems like something very unlikely given that we control the semantics of c:data documents. (The alternative is to create a p:decode step.)

### DIFF
--- a/steps/src/main/xml/specification.xml
+++ b/steps/src/main/xml/specification.xml
@@ -221,6 +221,7 @@ PSVI annotations.</para>
 <xi:include href="steps/compress.xml"/>
 <xi:include href="steps/count.xml"/>
 <xi:include href="steps/delete.xml"/>
+<xi:include href="steps/encode.xml"/>
 <xi:include href="steps/error.xml"/>
 <xi:include href="steps/filter.xml"/>
 <xi:include href="steps/hash.xml"/>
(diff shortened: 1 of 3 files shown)


xml-project (commented):
I think we shouldnt reuse XC0203 here. It is already used in p:http-request for an error constructing a multipart boundary.

xml-project:
@ndw Sorry, my fault. The spec uses XC0213, but I read XC0203. The later is an error in p:http-request.

ndw:
I went looking again in case I missed something and I found XC0052 in the error list, but I don't see it in the spec. It is in the `steps/payloads/request_body.xml` source, though. Now I'm wondering if there's part of the spec missing. I'll dig a little deeper tomorrow.

xml-project (approved):

ndw:
We already had a defined, but unused because it was part of `p:http-request` that we deleted, error code for unsupported encoding. I don't see any reason to make up new error codes if we don't have to, so I've changed `err:XC0213` to `err:XC0052`.

ndw:
See also: https://github.com/xproc/3.0-steps/pull/666

PR #670, opened 2025-02-06
Attempt to address the BOM

### LINKED ISSUE #668: Ignore the BOM
I think XProc steps should ignore (discard) the byte-order-mark (BOM) when loading text documents.

I don't think there's any justification for keeping it, it absolutely causes problems, and if a user really wanted it, they could load the document with a non-text content type, like `application/octet-stream`

ndw:
Fix #668 

This is a first attempt, I think it's reasonable as far as it goes, but we should look to see if there are more places where the BOM should be addressed.


gimsieke (approved):

xml-project (approved):
There is a typo in line 142 "suport"

Issue #672, opened 2025-02-24
Should [p:]message be an option?


ndw:
I was asked earlier today about how to make debugging messages appear only when a static `$debug` flag is true.

This works:

```xml
<p:identity use-when="$debug" message="Hi mom"/>
<ex:my-actual-step/>
```

But it's a bit clumsy and if `ex:my-actual-step` doesn't actually read the output of `p:identity`, there's the possibility that the message could be output after the step runs.

This would be nicer, in many ways:

```xml
<ex:my-actual-step>
  <p:with-option name="message" use-when="$debug" select="'Hi mom'"/>
</ex:my-actual-step>
```

It happens that I treat message attributes as a special kind of magic option (because that was the easiest way to implement the fact that they're AVTs) so XML Calabash allows this.

I don't think this is conformant with the spec, but should it be?

gimsieke:
DIY serializing, at least in Saxon, isn’t equivalent:

```xml
    <xsl:variable name="doc">
      <doc/>
    </xsl:variable>
    <xsl:message select="'serialize($doc):', serialize($doc)"/>
    <xsl:message select="'$doc:', $doc"/>
```
→
```
serialize($doc): &lt;doc/&gt;
$doc:<doc/>
```

xml-project:
@gimsieke Right:  `xsl:message/@select` is not a serialization, but is defined as a deep copy:

> If the [xsl:message](https://www.w3.org/TR/xslt-30/#element-message) instruction has a select attribute, then the value of the attribute must be an XPath expression. The effect of the [xsl:message](https://www.w3.org/TR/xslt-30/#element-message) instruction is then the same as if a single [xsl:copy-of](https://www.w3.org/TR/xslt-30/#element-copy-of) instruction with this select attribute were added to the start of the [sequence constructor](https://www.w3.org/TR/xslt-30/#dt-sequence-constructor).

If we leave `select` as xs:string for now and take the mechanism described above to XProc Next, we could easily changed the expected type to `item()*` without breaking anything, right?

gimsieke:
> ndw wrote: 1. Make the select attribute item()*, add a serialization option and say that the items are serialized for output. Potentially simpler for users, but also a bit more complicated. And also a bit more like xsl:message

A serialization option isn’t necessary. The processor can serialize an item as it deems appropriate. It can be a quality-of-implementation issue whether, for instance, a map that contains XML fragments deep down is serialized as `json` or `adaptive`.

[...]

PR #673, opened 2025-02-27
Describe a p:message step

ndw:
It seems like a consensus is forming to add a `p:message` step, so I took a stab at describing it.

The only open question in my mind is, if the `test` expression is `false()`, is the implementation forbidden from evaluating the select expression or is that implementation-dependent. (Or implementation-defined, I guess.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -101,7 +101,7 @@ task make_etc() {
 
 ["steps", "step-file", "step-os", "step-paged-media",
  "step-run", "step-text", "step-mail", "step-rdf", "step-validation",
- "step-ixml"
+ "step-ixml", "step-message"
  ].each { spec ->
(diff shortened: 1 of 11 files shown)


ndw:
Fix #672

xml-project (approved):

xml-project:
> It seems like a consensus is forming to add a `p:message` step, so I took a stab at describing it.
> 
> The only open question in my mind is, if the `test` expression is `false()`, is the implementation forbidden from evaluating the select expression or is that implementation-dependent. (Or implementation-defined, I guess.)

I am not sure we need to say something special here. There are other cases in evaluating a pipeline where a processor knows that an option is not used and therefore does not have to be evaluated, e.g. if match in p:add-attribute does not match anything, attribute-name and attribute-value don't have to be evauated.

The difference in user experience is (a) enhanced speed and (b) missing error messages resulting from the non evaluated expressions, right? I would think this could be implementation-dependent.

PR #675, opened 2025-02-28
Making select required in p:message

xml-project:


gimsieke (commented):
Maybe a bit late to the party and not exactly about this particular change:

In XSLT, I often put other items than strings (elements, whole documents, until Saxon 10 also attributes) into the `xsl:message/@select` attribute. If it only allowed strings, I’d need to take care of serialization, sometimes type casting and concatenation myself. 

What about accepting `item()*` for the `select`? And if we accept also an empty sequence, we need not make `select` required. It could be made implementation-defined what happens if the message content is empty: Does the processor write something like `Message: ` then or nothing at all?

ndw:
@gimsieke 's point is a good one. Sigh. Everything is more complicated than you'd like. Choices:

1. Make the `select` attribute `item()*`, add a `serialization` option and say that the items are serialized for output. Potentially simpler for users, but also a bit more complicated. And also a bit more like xsl:message
2. Leave it as it is, make users write `serialize()` themselves.

I want the answer to be 2, but I have a feeling that the right answer is 1.

ndw (approved):

xml-project:
@gimsieke Good point. For now using fn:serialize() should do a good job, doesn't it. So I would ask you to postpost this new feature to V next. Is this possible?

PR #676, opened 2025-02-28
Changed type of select

xml-project:
I tried to spell out @gimsieke 's idea: Is it as easy as this?

### DIFF
--- a/step-message/src/main/xml/specification.xml
+++ b/step-message/src/main/xml/specification.xml
@@ -89,7 +89,7 @@ A machine-readable description of this step may be found in
   <p:input port="source" sequence="true"/>
   <p:output port="result" sequence="true"/>
   <p:option name="test" as="xs:boolean" select="true()"/>
-  <p:option name="select" as="xs:string" required="true"/>
+  <p:option name="select" as="item()*" required="true"/>
 </p:declare-step>
 
 <para>Steps that produce messages offer one way for an author to keep track of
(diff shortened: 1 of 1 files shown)


ndw (approved):
I think that covers it.

ndw:
Yes. I think you're right.

I was going to argue that the name `select` and the parallel with `xsl:message` would strongly encourage people to believe that it should just be an expression and we should do it that way. Then I remembered that the static context for a step doesn't include the in-scope variables.

If we make it an expression, you can type `select="'test: ', ."`, but you can't type `select="'test: ', $varname"`. The work around for `.` is just use `p:with-option`, the work around for variable names is really ugly expression like

```
select="'test: &quot;' || $varname || '&quot;'"
```

Where you have to make sure that the variables are resolved while the string is being constructed but the resulting string is still a valid expression. Yuck.

xml-project:
@ndw So your idea is to go back to select as xs:string and wait for better ideas to come up?

ndw:
No, I think `item()*` is fine, it means I can put more complicated things in the output if I use `p:with-option`. I'm just abandoning my thought that it should be an expression resolved by the step.

xml-project:
Oh, now I understand what you had in mind. Haven't thought of it this way. Is this why it reads:

> If the test attribute is “true”, the select expression **is evaluated** and made available. 

Should we change this?

ndw:
I don't know. The `[p:]message` attribute speaks of computing the value, but it has special rules about `expand-text`.

Does this seem like an improvement?

> If the test attribute is “true”, the value of the select option is made available.

gimsieke (changes requested):
the result of evaluating the `select` option is serialized and made available

gimsieke (approved):

ndw (approved):

